PR #18, opened 2018-04-16
Add setup instructions for Windows to CONTRIBUTING

escherina:
The setup instructions for the device and hub simulator don't work with standard Windows terminals, so I've updated them in case any other Windows folks want to contribute. ✨


noopkat:
Thank you so much for this! I totally forgot that windows needs to use `set` for env variables.
Also I made a tonne of changes to running the dev environment because we'll need to transpile for Firefox's module support, so don't hesitate to open an issue if you see anything else broken! :bow:

bitmoji

PR #25, opened 2018-04-17
Add Background Image URL and Background Repeat fields to the settings - closes #6

### LINKED ISSUE #6: Background image tile feature
Similar to specifying a background color for the dashboard, it'd also be great to specify a url that points to an image to tile as the background of the dashboard.

We could call the new dashboard property `bgImageUrl`

+ [ ] add background image option to settings form (a text input would be fine)
+ [ ] add background image property to dashboard.json (leave as empty string)
+ [ ] ensure property is saving to dashboard.json when saved in settings
+ [ ] ensure that the dashboard doesn't break if background image property is an empty string
+ [ ] ensure background image repeats or 'tiles' across the whole dashboard if it's a small image

Following a similar code strategy to the background color setting would be helpful.

Any questions - comment on this issue 🙇‍♀️

escherina:
- Add inputs for Background Image URL and Background Repeat in the settings component template
- Make sure Background Repeat is always true by default in `dashboard.json.sim` and in the `helpers.js`
- Add some basic styling for the checkbox field and label. I tend to use BEM naming conventions for class names, so did that here.
- Update the `dashStyle` function to handle the new settings fields

A note about the if statement in the `dashStyle` function: I found that although using v-model with a checkbox sets the value to a boolean, the `saveSettings` function uses FormData, which stringifies booleans - so there was a chance of the `dashboard.bgImageRepeat` value being a boolean or a string at different times. To cover all bases, I convert it to a boolean in the `dashStyle` function. There may be a better way to do this...


noopkat:
solid PR 💯 

Pulled it down and played with it. Works great 💪🏼
The BEM styling is very 😍  

Sorry about the FormData bool issue. Perhaps using FormData was not as convenient as I thought it'd be 🤔

noopkat:
merged. Thank you so much! bitmoji

escherina:
bitmoji

No worries about the FormData issue - that's what I would have used too! I've not run across that problem with it before, but now I've learned something new. 👍

Issue #22, opened 2018-04-17
Simulation UI status callout


noopkat:
It would be nice to have a small status near top of the dashboard UI making it clear that you're simulating data when in simulating mode. I want to leave the Glitch app in simulation mode instead of hosting telemetry from a real IoT Hub, but make it clear that it's not 'real' data coming in.

When folks remix the app it won't start up in simulator mode so there's no worry about it persisting.

escherina:
I can take this one on. 😊

PR #26, opened 2018-04-18
Add simulation mode status message to app

escherina:
This PR creates a Simulation mode status as described in issue #22.

- It uses webpack's [DefinePlugin](https://webpack.js.org/plugins/define-plugin/) to expose the `process.env.SIMULATING` environment variable to the Vue application.
- If `SIMULATING` is `true`, a small status message is shown, situated inside the `<header>` but pinned to the top right of the page.
- Further enhancement might be some improved styling for the message. I considered putting it inside a bar with a solid background, but didn't want to clash with papayawhip❤️ or mess with the layout too much!
- I've tested that DefinePlugin works on Glitch with a [simple test project](https://glitch.com/edit/#!/vine-structure) that uses an .env file with `SIMULATING=true`.

### DIFF
--- a/public/css/style.css
+++ b/public/css/style.css
@@ -266,3 +266,11 @@ select {
   margin: 0 0 0 5px;
   width: auto;
 }
+
+.simulation-status {
+  font-family: 'Chivo', sans-serif;
+  font-size: 0.875rem;
+  position: absolute;
+  top: 5px;
+  right: 10px;
+}
--- a/public/js/components/app.js
+++ b/public/js/components/app.js
@@ -16,6 +16,7 @@ const template = `
   <div id="dashboard" v-bind:style="dashStyle">
     <header>
       <h1 v-bind:style="headingStyle"><span class="hemoji">⚡️</span>electric io</h1>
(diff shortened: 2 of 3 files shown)


noopkat:
@escherina this is 🔥 !

The styling and positioning is exactly how I imagined it. I wanted to keep it subtle as a small indicator.

I have one small request to make this perfect - would you mind applying the same [color contrast function](https://github.com/noopkat/electric-io/blob/master/public/js/lib/colorContraster.js#L8-L20) that I [use on the main heading](https://github.com/noopkat/electric-io/blob/master/public/js/components/app.js#L57) so it's still visible on dark (plain) background colours?

I am really enjoying all of these pull requests from you! 👏

escherina:
Oh, great point! I was just admiring that function the other day, too. 😄 I'll make that change later today, no problem.

And I'm really enjoying contributing! 👍

escherina:
Done! I've reused the `headingStyle` computed style to avoid repetition. If `headingStyle` gets additional styling in future, then the simulation status could get its own computed style. 😊

noopkat:
thank you! This is great. Merging now ✨ 

bitmoji

PR #32, opened 2018-04-28
Add text card to app

mpaarating:
Add a text card type to the app. This card uses the [marked](https://github.com/markedjs/marked)  package for standard and github flavored markdown.


noopkat:
Hi @mpaarating!

This feature is super nice, both code-wise and implementation wise! Thanks so much for such a stellar job 💯 

bitmoji

PR #33, opened 2018-04-28
[WIP] Fixes #7 form save status messages

### LINKED ISSUE #7: Status for form submissions
I'd like to see a text status appear on both the dashboard settings form and the card settings forms.

+ [ ] Dashboard settings - show both a success or an error.
+ [ ] Card settings - show an error below the save button if there's a save error, 'flip' the card as normal if save was successful

A good example to take lead from is the button card. When you click the button, it shows the status of the method call below the button.

### LINKED ISSUE #17: New card ideas
Comment below if you have any idea for new types of cards to have 💡

noopkat:
I programmed this rather feebly last stream because of an unfortunate personal distraction I had going on at the same time.

It's a pretty small feature but for a gut check I'd like some quick eyes on this to see if it could possibly be improved.
bitmoji


noopkat:
looking at #7 now I am missing the card saving part too for the card form. Will complete that and flag when this is ready for review

escherina:
Hey Suz, this looks great! I have a couple of ideas for the save status text, but they're minor - the code you've put together is very sound.

1. Currently the save status text element (`<p class="status">`) in the settings panel is creating some white space when there's no text visible, because it has a margin. Rather than showing the element all the time, it could use something like `v-if="saveStatusText"` to remove the element when the save status text is empty.

2. In the settings panel, the save status text persists after the settings are saved. It might be nice to have the text disappear, either after a timer has elapsed, or by allowing the user to dismiss it. I think the latter option would be better, as it might be easy to miss something on a timer and I don't think it's accessible. 
    - One way to do this might be: a dismiss button next to the save status text element could emit an event up to `app.js`, then trigger a method to reset the `saveSuccess` variable back to `null` (then the computed property `saveStatusText` will get set to an empty string).
    - We could do the same thing with the save status text on the cards, although I think it's needed less there - the current flow of showing error text when it fails and quitting editing mode when it succeeds seems good to me.
    - This point is pretty minor, so let me know if I'm over-thinking this! 😅 

Let me know if you'd like me to add a commit to this PR with either/both of these suggestions. 👍

mpaarating:
@escherina ignorant question that probably doesn't apply to an app this small, but I thought I remembered reading that `v-if` isn't very performant.
A). Have you noticed that at all in your practical experience?
B). If yes, should the class instead be changed to a `display: block`/`display:none`?

escherina:
@mpaarating That's a great point! The Vue apps I've worked on have used both `v-if` and [`v-show`](https://vuejs.org/v2/guide/conditional.html#v-show) (which toggles the `display` property under the hood) and I must admit, I haven't noticed a difference. (Caveat: the elements being toggled were never very complex, so maybe not the kind of situation where you would notice a difference.) There were also some situations where we wanted to remove the elements from the DOM, so had to use `v-if`. I'd like to go back and test it more thoroughly now! 😄

They address the question of `v-if` vs `v-show` [in the Vue docs](https://vuejs.org/v2/guide/conditional.html#v-if-vs-v-show) and have some recommendations:
> Generally speaking, `v-if` has higher toggle costs while `v-show` has higher initial render costs. So prefer `v-show` if you need to toggle something very often, and prefer `v-if` if the condition is unlikely to change at runtime.

In this case it doesn't seem clear-cut as to which we should use, as the save status text shouldn't be toggled "very often". However, erring on the side of caution, I'd say we should go with `v-show` here.

Thanks for the thoughts - I hadn't really looked into the `v-if`/`v-show` comparison before, and I am now more enlightened. 👍

noopkat:
thanks for all of the feedback - I only just saw this (oops) and it's 💯 to have some improvements to now go and make to this PR!

I'll solve the merge conflicts and look into the `v-show` / dismissible statuses.

bitmoji

noopkat:
closing this for now because it's too out of date with the latest changes to merge it. Will revisit soon!

PR #34, opened 2018-04-29
Added jest and a test test

underscoredotspace:
I've amended `.babelrc` so that Jest is correctly handling the es6 modules per [the Jest docs](https://facebook.github.io/jest/docs/en/webpack.html#using-with-webpack-2), hopefully this isn't breaking anything else (I don't think it should). I've not used Webpack in a project yet, so I have no experience with this bit. 

Also, I know coverage is being generated on files it shouldn't, so currently it's taking a while to do that. If you want to let me know which files should be covered I'll update the pull request. 😁

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,7 +1,12 @@
 {
+  "presets": [["es2015", {"modules": false}]],
+
   "env": {
-		"presets": [
-			["env"]
-		]
+    "presets": [
+      ["env"]
+    ],
+    "test": {
+      "plugins": ["transform-es2015-modules-commonjs"]
+    }
   }
 }
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 node_modules
+coverage
 .data/dashboard.json
 .glitch-assets
 .DS_Store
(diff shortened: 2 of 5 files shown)


escherina:
This looks fab, @underscoredotspace! I love a good Jest setup - that built-in coverage is 💯.

In terms of the coverage, I think you can safely exclude the vendor files (`public/js/vendor`) and the compiled application (`public/js/dist`), which should improve the speed.

I'll leave this open for now in case @noopkat has any other files she wants excluded.

bitmoji

underscoredotspace:
Thanks @escherina. Based on that, I reckon the below should do it: 

```
collectCoverageFrom": [
  "lib/**/*.js",
  "public/js/**/*.js",
  "!public/js/{dist,vendor}/*.js"
]
```

underscoredotspace:
I've had a good dig through the code and the above seems right, so I've updated the PR. Hopefully it's good to go

mpaarating:
As an aside, this looks like a great addition to Jest. I've not had a chance to actually use it, but chained assertions seem 💯 https://github.com/mattphillips/jest-chain

atbuzz:
Great idea!

escherina:
I've finally had a chance to revisit this, apologies for the delay! Everything looks good to me, so I'll merge this in now.

bitmoji

PR #37, opened 2018-05-23
Simplify Form Component

### LINKED ISSUE #10: Simplify card settings form component

mpaarating:
The thinking here is that the template should be kept with all other form related default settings. We can avoid nasty switch and other control flows by doing it this way.

I do see an issue where someone will need to regenerate their `dashboard.json` in order for this to work.

Closes #10


mtheoryx:
In general, this looks pretty good. I only have one concern. It looks like the "starter" dashboard used for simulation has been removed. I may be overlooking something, but that file was used in simulation mode as a quick boostrap for quickly getting spun up.

mpaarating:
Good call. In addition to that, I'd like to provide progressive enhancement support new features on started (basically parse a dashboard.json file for missing keys and fill in with null values), I'll open an issue for that though.

mpaarating:
@mtheoryx I got the sim file added back in and also added the `public/js/dist` folder to the `.gitignore`. Seemed silly to keep it in there. Any other feedback?

mtheoryx:
I think we're looking good, nice work!

noopkat:
@mpaarating this is *fantastic*. I'm going to pull this down tomorrow and manually test it very quickly, but I love it.

thanks @mtheoryx for your feedback / review. Super helpful.

Love both of your contributions to electric-io. Y'all are making my week.

noopkat:
hello @mpaarating hey! Thanks for the feature. We reviewed this on my Twitch stream and it looks great 👍  Looking forward to the progressive dashboard updater / backfiller 😄 

bitmoji

Issue #11, opened 2018-04-16
Support hiding of settings pane


noopkat:

escherina:
I can take this one! @noopkat - what would be your preferred way to hide the panel? I was thinking to collapse it, like an accordion. Since it's pinned to the right hand side of the screen by default, it could collapse into a horizontal bar, a vertical bar - or something smaller.

I put together a few options [in Figma here](https://www.figma.com/file/NBSU5KlAwKxDgp0CPbw9aLdJ/Electric-IO-settings-panel-options). I think the settings icon version (option 3) might be good for mobile, but otherwise the horizontal bar version (option 1) seems the most obvious for the user.

noopkat:
Hi @escherina!

Thank you so much for the Figma mockups! Super helpful. Let's go with Option 3 (the gear!) 💯

PR #42, opened 2018-05-27
Add show/hide functionality to Settings pane

escherina:
Addresses #11: Support hiding of settings pane.

I've added a toggle button to the Settings pane that allows the user to show and hide the card at will.

**Accessibility**
- I'm using `aria-expanded` to let screen-readers know if the pane is expanded or collapsed.
- The button is labelled with `aria-labelledby` which points to a screen-reader-only element inside the button, which reads either "Close settings panel" or "Open settings panel".
- The icons are completely hidden from screen-readers with `aria-hidden`.
- For keyboard-only users, the title of the pane ("Settings") appears when the button is focused. It also shows on hover. This does mean that immediately after closing the pane, the title is shown until the user clicks/tabs away, as the button is still focused.

**Note on animation**
It would be nice to animate the opening and closing of the card, and I considered animating the size of the pane using the `scale()` transform, but this would also animate the scale of anything inside the card. There are ways to work around this, but rather than putting a lot of effort in, I thought it would be better to discuss animation and to add it later if needed.

**Minor style refactor**
I've removed `id="settings"` from the pane and instead moved everything into class-based styling, to make it easier to override in future. I've also moved all the settings-related styling into its own section in the `style.css` file. I've added a couple of wrapper `<div>` elements and some additional classes inside the template t

**Other thoughts:**
- The styling may need some work - e.g. I've left the original button focus style (2px black border) in place.
- For the icons, I've used the gear emoji (⚙️) and the `&rarr;` right arrow symbol (&rarr;). However, we could use svg icons, which might look better. We could also use a &times; symbol rather than an arrow - but since the pane isn't completely dismissed, I thought that might be confusing.


mpaarating:
This is fantastic work, especially with the focus on accessibility and browser support. 

As the `settings.js` file has grown a bit out of this, I think it might make sense to create a ticket to refactor the settings form into its own file (I'd be happy to do it). Thoughts? (I tend to go overboard with thin files).

noopkat:
@escherina sorry for the delay in pulling this down. I am very excited to play with this hopefully tomorrow, and as usual thanks for keeping a11y in mind 🌟 

@mpaarating makes a good point. I'd prefer to merge this tomorrow if all looks good, and then create a ticket for the settings refactoring so this can be released sooner 🙇‍♀️

noopkat:
@escherina could you please merge master into this branch and push back up? I'm nervous I'm going to possibly break your code in subtle ways if I do it myself 🙇‍♀️ 

Thank you!

bitmoji

escherina:
Merging this in now - it was primarily whitespace, but also needed to add in the new App Title field.

bitmoji

noopkat:
bitmoji

Issue #13, opened 2018-04-16
Document how to migrate dashboard settings from one app to another


noopkat:
This should go in the README.md file.

Essentially to migrate your dashboard layout settings what you do is copy your dashboard.json file from one electric-io app to another, placing it in the `.data` directory.

davidk:
Hello there,
Will give this a go!

davidk:
Issue #53 (Windows specific) is now up for grabs, so this is a friendly wave to close this one :star2:

noopkat:
@davidk 🙌

PR #46, opened 2018-06-03
README.md: Add a section on transferring dashboard settings

davidk:
Hello and greetings,

This PR addresses issue #13. It seems possible to transfer the configuration live, but I believe it'd be best if a user stops both instances so that the configuration remains in a consistent state.

Addendum: I wanted to be sure that I didn't break any user trying this, so I spun up an impromptu
Docker container that runs two electric-io instances in parallel. Instructions for fully reproducing and tearing down are embedded in the Dockerfile here:

Dockerfile: https://gist.github.com/davidk/059e10b6fc04aa02bea058c7dfda66b8

Addendum 2: As I am uncool and without bitmoji, I submit this extremely friendly :cat2: in its stead. Spotted in Berkeley, California (if by kismet you meet it, say hi over and over in the dorkiest way possible to be greeted with a boop):

a cat lounging in the sun

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,6 +118,17 @@ Fields:
 
 Got ideas for more cards? Open an issue on this repo and let me know! 👀
 
+## Transferring dashboard settings
+
+electric-io saves your dashboard configuration to `./.data/dashboard.json`. To safely transfer this elsewhere, stop the source and destination electric-io instances and copy `dashboard.json` across.
+
+```
+# With both the source and destination electric-io instances stopped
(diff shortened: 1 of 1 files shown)


davidk:
@mtheoryx Not at all! Just a disposable Dockerfile for testing out this particular PR's behavior

noopkat:
@davidk this is awesome, thank you so much for testing this approach so extensively with Docker! 🎉 
Also, that cat picture totally made my day; what a cutie!

bitmoji

Issue #40, opened 2018-05-27
Clarify supported NodeJS version


mtheoryx:
From the readme, it is unclear what version of Node is required for the project.

I propose:

1. Add the supported version to the readme
2. Add the version in package.json under "engines" and "enginesStrict"
3. Using the correct base image for Docker.

I'm working on task 3 now... I really just need a definitive decision from the maintainers on where our base node version will start.

I'm happy to PR this once we make a decision.

Cheers!

escherina:
Great idea! In the [contributing guide](https://github.com/noopkat/electric-io/blob/master/CONTRIBUTING.md), it says that NodeJS 4+ is required. However, the app will ultimately run on Glitch, which, [looking at their FAQ](https://glitch.com/faq#node), only goes down to version 6.0.0 (https://nodeversions.glitch.me/) - so maybe we could set 6.0.0 as the minimum version? What do you think, @noopkat?

mtheoryx:
Interestingly, in the package.json, we currently have

```json
	"engines": {
		"node": "8"
	}
```

So I'm not sure if 4 was the target, or where the 8 came form.

I think, like browsers, "current + (N - 1)" could work. So active LTS is 8.x "carbon" and Maintenance LTS is 6.x "Boron" https://github.com/nodejs/Release#release-schedule

What do you think? Current, plus one version behind or maintenance?

mtheoryx:
I included the node version updates and clarification in my PR. I needed to specify a known good node version for the base Docker image anyway :) I'll close this issue!

PR #47, opened 2018-06-03
Add option to use containerized app with Docker

### LINKED ISSUE #44: Support Docker for isolation, portability, and quick setup
Opening this (and working on it) for the following reasons:

1. Isolation - no need to pollute your global with different versions of "stuff"
2. Portability - reduce "works on my machine" weirdness across environments
3. Quick setup - Minimize install requirements to quickly spin up an example

mtheoryx:
There's a couple of things going on with this PR, so I want to lay it out for anyone unfamiliar with Docker.

**Why?**
- Isolation. You don't even need to install node on your machine.
- Dependency control. Like package.json, even the base OS and node version are specified in code.
- Environment parity and portability. You will get the exact same running container no matter what OS you or your contributor are using.

The Dockerfile, think of it like ingredients for making cupcakes!

This is a line-by-line set of instructions to build an application image. This image can be versioned and pushed to a Registry, allowing others to use it, or even build on top of it.

The docker-compose.yml file. This is your instructions on when to add ingredients, and how to bake and ice the cupcakes!

**A few things to note!**

Because containers are stateless, we are mounting the application code from your machine (the host) into the container. When things change in the container... they are persisted on your host. 

The SIMULATING mode is still valid! You can either use the .env file, or you can pass in an override to turn simulating mode on. I updated the readme for this, too, but it looks like:

```bash
SIMULATING=true docker-compose up --build
```

Please feel free to try the new instructions in the README, and let me know how it works for you! And ask any questions right here in the PR, I'd be happy to explain/answer them.

This should close #44 and should address the node version concern in #40


.dockerignore
```diff
@@ -0,0 +1,3 @@
+node_modules
```
mtheoryx: This just saves space in the docker container as Docker does not care about these files/directories.

docker-compose.yml
```diff
@@ -0,0 +1,20 @@
...
+  app:
+    build: .
+    image: electic-io
+    container_name: electric-io
+    command: npm run start
+    ports:
+      - 3000:3000
+    env_file:
+      - .env
+    environment:
```
mtheoryx: This allows the environment override for simulating. I could expand this for any of the .env options if we need to. I figure people can modify their own .env files for the other settings and not check them in if they are contributing, but toggling simulating seems like something you'd want to do more often.

docker-compose.yml
```diff
@@ -0,0 +1,20 @@
...
+      - SIMULATING
+    volumes: # For file editing, watching, etc
```
mtheoryx: This preserves the changes made in the container back to your host. And also changes (as you're developing) on your host get reflected back into the running app container.

package.json
```diff
@@ -20,8 +20,9 @@
 		"socket.io": "*"
 	},
 	"engines": {
-		"node": "8"
+		"node": ">= 6"
```
mtheoryx: Per discussion in #40 we settled on a minimum version of 6. I do know that 8.x works as well, so went with >=.

noopkat:
@mtheoryx this is an amazing first pull request - thanks so much! I'm excited about the potential for this project now being container compatible. Let's chat about putting this on Docker Hub asap 💯 

bitmoji

PR #56, opened 2018-06-10
fixes #54 adds debug package for firehose logging

### LINKED ISSUE #54: Environment option for debug/firehose?
Suz turned off the firehose in simulation mode, but also mentioned a debug mode possibility?

We can really easily support that with env vars.

Thoughts?

[maintainers: a good label would just be "question" or something]. Cheers!

noopkat:
👋 hello!

Kinda does what it says on the tin? 👀 
I decided to add the debug package because I really like the features it offers.

Thanks for the help on the stream everyone! 💖 

bitmoji

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -12,6 +12,7 @@ services:
       - .env
     environment:
       - SIMULATING
+      - DEBUG
     volumes: # For file editing, watching, etc
       - ./public:/usr/app/public
       - ./lib:/usr/app/lib
--- a/index.js
+++ b/index.js
@@ -8,6 +8,7 @@ require('dotenv').config();
 var express = require('express');
 var fs = require('fs');
 var http = require('http');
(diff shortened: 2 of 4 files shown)


mtheoryx:
Fantastic!

@noopkat one think that I know we've been kinda mentioning is documentation of the .env and potential ways to use those env vars in docker, and just the CLI.

I wouldn't want to hold this PR up, so maybe we could have a standalone issue around "clarify documentation around environment variables and the environment file?"

It would consolidate where changes happen (probably just in the readme) and also be a candidate for "good first issue".

What do you think?

mtheoryx:
Just for clarification, I think this one is good to go!

noopkat:
yes agreed - let's raise an issue for the docs! ✨

mtheoryx:
New issue created #60

Issue #58, opened 2018-06-11
Support for video streams


strangecyan:
I’d love to have support for video streams within a card from a home automation system or a raspberry pi camera. I’m unsure what kind of config you’d need under the hood, whether a url to point the browser to would be enough or if we need more information about the configuration of the camera?

mtheoryx:
That's a really good idea! I have no idea how to implement that.

I'd also say maybe we'd still want some "sample video" for use in the demo dashboard or simulating mode.

Anyone know much more about video stream options? @mpaarating @noopkat

mpaarating:
@noopkat correct me if I'm wrong, but I don't believe that IoT hub itself provides video streaming services. (Not really sure any of the IoT centers do) We _could_ add a HTML5 video card with a spot to paste in an http stream.

noopkat:
I LOVE THIS IDEA!!

Ok, so we could support a couple of video stream types here:

1. Video element
2. Image element for jpeg based streams

the sticker card would probably already support the jpg streams, so the video element would be all that's necessary.

> We could add a HTML5 video card with a spot to paste in an http stream. Would that be useful?

@mpaarating ^^ Yep! 😄

noopkat:
I researched this a little today and HTML5 video elements still don't support live streaming video very well. An additional JavaScript library will be needed to get this working.

MJPEG streams will currently work by using a sticker card.

noopkat:
closing this as we have mjpeg support right now, but feel free to reopen if you want to see other video formats.

PR #63, opened 2018-06-28
Add MJPEG stream note

MichaelE1:
In light of https://github.com/noopkat/electric-io/issues/58 - I thought it might be worth mentioning the possibility of using a MJPEG stream in the docs :blush:

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,7 +157,7 @@ This card will display a sticker on your dashboard.
 Fields:
 
 1. Title - text displayed at the top of the card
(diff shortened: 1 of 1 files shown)


mtheoryx:
I second that request, just imagine someone knows nothing and throw some "launching off points" in the docs, and that would be awesome!

mpaarating:
Looks great! Thanks for the extra input and totally agree about being generally vague. Great work!

Issue #59, opened 2018-06-12
More heavily leverage Webpack for a more standard Vue setup


mpaarating:
E-io has been a fantastically simple project for some time now, but I think we might be entering a point when using `.vue` files, transpilation, and a more traditional webapp setup (bring Vue in via `package.json`) would help developer efforts.

The testing issue is certainly an issue that seems to be tied to the way that Vue is imported in the project (I'm sure that there are ways around this and that someone can figure it out). Using `.vue` files would also give some useful syntax highlighting for those new to Vue and might help to lower the contributing barrier to entry.

Thoughts?

mtheoryx:
I certainly see the issue with testing because of how we're loading in Vue. I think you may be onto something here @mpaarating 

I'm certainly not a Vue expert, but it's worth asking: Can we get away with fixing the testing issue without a wholesale change to making this an idiomatic "Vue app"?

And if so... what other issues down the road are we still going to be dealing with? Is going "Vue app" an inevitability?

underscoredotspace:
The problem we had with testing the Vue components has been resolved, thanks to @escherina finding the missing piece, so I feel like this is _less_ of a barrier. 

This said, it would be easier for Vue newbies like myself to contribute more if the components were more idiomatic and matched what is shown in the documentation.

mtheoryx:
Fantastic @escherina and @underscoredotspace !

mtheoryx:
@mpaarating What do you think the first steps are? I think you're way more familiar with Vue than I am. Our mutual contact, Jeff, isn't listed in the Forks so I don't think he attempted a "conversion".

mpaarating:
So I've not had a ton of time to look at it yet, but my thinking is to convert the vue component files to `.vue` files and then transpile/concat/minify. As its a day off, I'm hoping to get a good pass at this today.

noopkat:
> my thinking is to convert the vue component files to .vue files and then transpile/concat/minify

Yep! This is what we need to do. It's not very glamorous work, but would improve the codebase a lot 😄 especially for potentially new contributors who are used to the more idiomatic vue way.

mpaarating:
Made really good progress with this today. I got most of the app converted to `.vue` files, but ran into an issue late in the evening. I'm sure sleep will solve the issues.

noopkat:
that's great. I watched part of your last stream and was grinning from ear to ear 😄 
this issue's work is a bit of a chore so I really appreciate all of the hard work 💯

[...]

PR #64, opened 2018-07-25
Modify webpack for transpiling

mpaarating:
This PR is in reference to #59. The biggest changes here are in the `webpack.config.js`. There are a lot of opinion things that occur in this PR (mostly renaming things), so if no one likes the names, let me know and I'll revert them! 

There's some other things that need done such as adding a "production" build with minification etc.

Thanks for all of your patience while I found time here and there to work on it!


package.json
```diff
@@ -1,65 +1,87 @@
 {
```
underscoredotspace: The white-space change in this file makes it difficult to see the functional changes. I know we discussed getting a prettier pre-commit hook for future, but maybe you could fix this one before merge?

underscoredotspace:
I don't have bandwidth to do a good dig through this (job hunting/code tests/bootcamp graduation 😓), but at first blush it looks good. I'll come back and add more comments if I have time later though.

package.json
mpaarating: I can totally do that. As a possible interum solution, could you see if the `hide whitespace changes` option in diff settings helps at all? Sorry for all the white space/formatting changes.

image

package.json
```diff
@@ -1,65 +1,87 @@
...
+  "scripts": {
+    "start": "npm run watch",
+    "watch": "npm-run-all --parallel watch:server watch:build",
+    "watch:build": "webpack --progress --colors --watch",
+    "watch:server": "nodemon index.js",
+    "test": "jest",
+    "test:watch": "jest --watchAll"
+  },
+  "dependencies": {
+    "azure-iothub": "^1.2.5",
```
mtheoryx: We don't have to change this, but I'm a bit leery of having carrot-prefix on dependency numbers. I know that's the default when you `npm install` things, but... bad things can happen sometimes.

public/index.html
```diff
@@ -1,27 +1,24 @@
...
+  <link href="https://fonts.googleapis.com/css?family=Chivo|Nanum+Pen+Script" rel="stylesheet">
+  <meta name="viewport" content="width=device-width, initial-scale=1">
+  <link rel="stylesheet" href="css/style.css">
+</head>
+
+<body>
+
+  <div id="app"></div>
+
+  <script src="js/dist/vendors~main.bundle.js"></script>
```
mtheoryx: Is that `~` in the filename?

public/js/components/BaseCard.vue
```diff
@@ -32,13 +57,17 @@ export default Vue.component('card-base', {
...
-    }
+    };
   },
   methods: {
     onMouseDown: function(event) {
       const allowedModes = ['unlocked', 'demo'];
       const excludedNodes = ['INPUT', 'TEXTAREA', 'SELECT', 'LABEL'];
-      if (!this.dragging && !excludedNodes.includes(event.target.tagName) && allowedModes.includes(this.editMode)) {
+      if (
+        !this.dragging &&
```
mtheoryx: Excellent, thanks for splitting this up. Great example for the future!

public/index.html
mtheoryx: Disregard! Wasn't familiar with that notation!

mpaarating:
So I had joked about the line chart card working on my machine, but it literally does. Could others confirm that it doesn't/does work on their machines? 

[Image from Gyazo](https://gyazo.com/fc42272d74a926647ff08fab694f504b)

package.json
mpaarating: I totally agree. I'll lockdown versioning a bit better before this gets merged.

mpaarating:
No worries! I should have suggested that earlier and it should really be part of the testing steps that a contributor provides in the PR (I should have included some sort of testing steps also). I suppose a PR template might be helpful in the future :D

package.json
underscoredotspace: do we think this kind of change belongs in this PR? caret deps exist prior to this. I feel like we're good to merge.

mpaarating: Agreed, I'll put up another PR in the next day or so to address the versioning in this file.

PR #65, opened 2018-08-02
Lockdown packages versions

mpaarating:
Changes package.json lock down dependency versions and allow for patch changes for dev dependencies.

Changes Express version to 4.16.x to allow for usage of bundled middleware such as `express.json`.


package.json
```diff
@@ -35,32 +35,32 @@
...
+    "vue-loader": "~15.2.0",
+    "vue-template-compiler": "~2.5.16",
+    "webpack": "~4.16.2",
+    "webpack-cli": "~2.0.14",
+    "webpack-watch-server": "~1.2.1"
   },
   "jest": {
     "moduleNameMapper": {
-      "^vue$": "vue/dist/vue.common.js"
+      "~vue$": "vue/dist/vue.common.js"
```
underscoredotspace: This looks like an oops?

mpaarating: Good catch, fixed 👍

mtheoryx:
I think this is a really good start. The `~` is certainly more restrictive than the `^`.

That said, I still feel like the `~` leaves open the possibility that one of the downstream dependencies changes, and ends up affecting the app.

Locking it down explicitly, as in `--save-exact`, with no save prefix is the safest, although maybe not the most convenient.

@mpaarating How do you feel about making the direct dependencies exact (no prefix) and making the dev deps looser? I've done that before.

package.json
```diff
@@ -13,16 +13,16 @@
...
-    "express": "^4.12.4",
-    "marked": "^0.4.0",
+    "azure-iothub": "~1.2.5",
+    "azure-iothub-receiver": "~2.0.0",
+    "babel-runtime": "~6.26.0",
+    "chart.js": "~2.7.2",
+    "dotenv": "~5.0.0",
+    "express": "~4.16.0",
+    "marked": "~0.4.0",
     "socket.io": "*",
```
mtheoryx: I'm not 100% sure why just allow star on this one.

mpaarating: So I've had some oddness with this one where I've used `2.1.1` and its crapped the bed (I wanna say that it got stuck on `UWS`. I can play around with it again as I didn't spend _much_ time trying to sort that one.

mpaarating:
@mtheoryx That's a good call. I wasn't sure how loose @noopkat was wanting to leave any of the dependencies.

noopkat:
merging

bitmoji

Issue #43, opened 2018-05-27
Add .editorconfig file to project


escherina:
Having a [.editorconfig file](http://editorconfig.org/) might be helpful for contributors whose IDEs can auto-format on save (like mine! 😄).

.editorconfig supports the following settings:
- `indent_style`: `tab` or `space`
- `indent_size`: e.g. `2` or `4` spaces
- `end_of_line`: `lf`, `cr`, or `crlf`
- `charset`: set to `latin1`, `utf-8`, `utf-8-bom`, `utf-16be` or `utf-16le` to control the character set.
- `trim_trailing_whitespace`: setting to `true` will remove any whitespace characters preceding newline characters
- `insert_final_newline`: setting to `true` ensures that files end with a newline when saving

Looking at the project files, it seems that these are the formatting settings:
- `indent_style`: `space`
- `indent_size`: `2`
- `end_of_line`: `lf`
- `charset`: `utf-8`
- `trim_trailing_whitespace`: `true`, I think?
- `insert_final_newline`: `true`

mtheoryx:
Interesting idea. I have mixed feelings about some of these things (more below, and would love to hear your thoughts), but whatever the project thinks is the best way forward, I can get behind :)

- A standard has to be decided upon, and usually, that's a team thing
- Linters are good, but can be annoying (we have a callout that we aren't using a linter here)
- Auto-formatters can be equally annoying when it runs live as you're trying to just work something out
- Alternatives are git hooks (pre commit, pre push, etc) to format code that may be less invasive

escherina:
Thank you for the thoughts, @mtheoryx! I agree with you that linting and anything that invasively changes code would be out of place for this project.

However, a .editorconfig file is more static than a linting setup, more of an convenience for contributors: if your IDE doesn't have an auto-format on save setting, or doesn't recognise .editorconfig files by default, the .editorconfig file won't do anything. And for contributors whose IDEs are set up to use it, it helps avoid extensive commit changes caused by accidentally reformatting project files. Its settings are to do with whitespace/line endings only.

There's definitely a conversation to be had around standards! 😄 In this case, I would say we wouldn't need to use all the settings. The most important ones in terms of formatting are the indent style/size and the line endings. I'm happy with any indent style/size, but since this is @noopkat's project, the existing indent of 2 spaces seems fine. There also seems no pressing reason to change the line endings from LF.

[...]

PR #66, opened 2018-08-02
Add prettier pre-commit hook

mpaarating:
Per #43, this PR adds a pre-commit hook to run the [Prettier default formatting](https://prettier.io/docs/en/options.html). Please give feedback on whether or not we need to the default settings and please provide suggestions! Thanks all!


noopkat:
thank you as always for the hard work @mpaarating ! ✨ 

bitmoji

noopkat:
:shipit:

PR #67, opened 2018-08-05
Fix case sensitive require paths (linux)

MichaelE1:
Hello!

I ran into some issues when trying to run the wonderfully webpack-ed version on my Ubuntu machine.

`ERROR in ./public/js/main.js
Module not found: Error: Can't resolve './components/App' in '/home/michael/electric-io/public/js'
 @ ./public/js/main.js 2:0-35 7:13-16`

and 

`ERROR in ./public/js/components/CardForm.vue?vue&type=script&lang=js (./node_modules/babel-loader/lib!./node_modules/vue-loader/lib??vue-loader-options!./public/js/components/CardForm.vue?vue&type=script&lang=js)
Module not found: Error: Can't resolve './FormFields' in '/home/michael/electric-io/public/js/components'`

I'm not a webpack expert but my understanding is that on MacOS the require paths are case insensitive but on Linux they are sensitive. In this PR I've renamed the offending files and this has solved the issue for me. 

I understand that some repos use the [case-sensitive-paths-webpack-plugin](https://github.com/Urthen/case-sensitive-paths-webpack-plugin) to help avoid this issue. But I haven't included it in this PR as it's probably something to that should be discussed first.


mpaarating:
Doh! I totally thought I had fixed that as I'd noticed a similar thing when testing on Linux. Sorry about that! Thanks for fixing it!

noopkat:
bitmoji

PR #69, opened 2018-08-27
Try and parse strings to numbers

burkeholland:
By default, the number card throws an error if a string is passed as the data field. We can use `parseFloat` to try and get at the number if it's a string.


noopkat:
thanks for this fix @burkeholland!

this is a *nice* guardrail indeed. Merging.

let me know if you spot any more improvements - we welcome them.

bitmoji

PR #75, opened 2018-09-18
Fix broken docker build

### LINKED ISSUE #74: Docker build broken and npm package mismatch potential
**TLDR**: I already have a fix for this, but wanted to record this correctly as an issue. I'll take it asap.

## Docker node_modules are just a whole buncha messed up

This is my bad. What's happening currently is that on build-time, docker seems to be working fine. However, on run-time, where the volume mapping comes in, we end up with new-hotness from several PRs ago not being reflected. And this is not 100% reproducible, depending on whether you already had this stuff installed on your dev machine.

Fix is to correctly map app and node_modules volumes in docker-compose, in addition to copying in the package-lock.json into the container for the new build step.

## Version mismatch between VUE and vue-template-compiler

Vue was at `2.5.16` as a dependency and vue-template-compiler was added at `~2.5.16`. The latter issued a patch fix `~2.5.17` and appeared to be causing an error.

Fix is to lock down vue-template-compiler to `2.5.16` to keep in step with vue releases.

mtheoryx:
So... I broke the build :|

image

2 things happened.

1. I was handling volume mounts irresponsibly and got into a super weird space after some updates. And I could reproduce it (yay). That's fixed now.
2. There was a patch version update of vue-template-compiler that caused "whelp guess I'll just give up and die" from the webpack dev server. I locked that down.

It was not locked down because it's generally safe enough as a devDependency... but these things can still happen!

So, we should be good!

image

Fixes #74

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@ FROM node:6.14.2-alpine
 WORKDIR /usr/app
 
 # Install app dependencies
-COPY package.json .
+COPY ./package.json ./package-lock.json ./
 RUN npm install --quiet
 
 # App source
@@ -12,3 +12,6 @@ COPY . .
 
 # Ports
 EXPOSE 3000
+
+# Default Command
+CMD ["npm", "run", "start"]
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -6,16 +6,14 @@ services:
     image: electic-io
     container_name: electric-io
     command: npm run start
-    ports:
+    ports: # Can override if you have port collisions with something else
       - 3000:3000
-    env_file:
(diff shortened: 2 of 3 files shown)


toolboc:
@mtheoryx , looks like I stumbled across the same issue in #78 , I fixed by adding as a regular dependency. in #79

noopkat:
hey @mtheoryx this seems chill 👌

toolboc:
@mtheoryx , I pulled changes locally and spun up a container instance with no issues, looks good!

mtheoryx:
@toolboc Thank you so much for the independent verification!

image

PR #79, opened 2018-09-18
fix vue-template-compiler error

### LINKED ISSUE #78: Error: vue-template-compiler must be installed as a peer dependency
Running from source using `npm start` without dev dependencies installed or attempting to run the docker image (which lacks dev dependencies)  results in:
`vue-template-compiler must be installed as a peer dependency`

image

Oddly enough, the app appears to operate fine when this error is thrown.

Looks like this can be fixed by adding `vue-template-compiler` @ same version of `vue` to package.json

*PR incoming

toolboc:
* fixes #78

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,8 @@
     "marked": "0.4.0",
     "socket.io": "2.1.1",
     "vue": "2.5.16",
-    "vue-focus": "2.1.0"
+    "vue-focus": "2.1.0",
+    "vue-template-compiler": "2.5.16"
   },
   "engines": {
     "node": ">= 6"


mtheoryx:
This does fix it in a similar way as #75 !

On separating devDepedencies, I see what you mean @noopkat and I generally would do that. However, I got curious and ran across this issue where Dan Abramov talked about why they moved something into dependencies:

https://github.com/facebook/create-react-app/issues/2696

tl;dr As opposed to something like test runners and assertion libraries, vue-template-compiler is required at build time. Their argument over on that issue was that some deployment providers treat it such that only 'dependencies' are needed for a build, and they were running into issues.

So, I'm totally on the fence about whether we do or not! If we want to do this, I can update my #75 quickly if we merge this. @mpaarating Thoughts?

mpaarating:
My gut reaction is that it should stay in devDependencies as the deployment issue that Mr. Abramov (blessed his name) referenced _shouldn't_ be an issue in this project. With that said, I don't have that strong of feelings either way.

toolboc:
I was able to verify that the changes from @mtheoryx in #75 provide the same error relief.  I say we go in favor of that PR due to keeping everything in devDepencies along with the additional changes it brings.  The duplication was due to my not noticing that a PR was already in place, however, we now have independent verification that @mtheoryx 's changes are good to go!

*closing

mtheoryx:
@toolboc No worries, kind of a funny coincidence we both stumbled on this at the same time Thank you for also trying to track it down!

PR #80, opened 2018-09-21
Azure support

toolboc:
This PR adds support for deployment of docker-compose.yml to an Azure container Web App running on a Linux App Service.

Summary of changes:
* docker-compose.yml sets environment variables directly in the docker-compose file to support deployment to cloud services without need for an .env file.
 * docker-compose.yml is set to pull the latest image from noopkat/electric-io which ensures that the latest automated build from dockerhub is retrieved instead of building from source.  This compromises building from source in favor of seamless deployment to cloud providers.  Please feel free to share thoughts on this approach.
* There is a documented issue where [socket-io must be instantiated with `perMessageDeflate` set to *false* in order to allow WebSockets to work in a Linux App Service on Azure](https://docs.microsoft.com/en-us/azure/app-service/containers/app-service-linux-faq#language-support)

Here is an example of deploying to Azure with these changes:
image

And here is a live site that is reading data from a particle photon using the included docker-compose.yml with only modification of the `CONNECTION_STRING`: https://electric-io.azurewebsites.net/

image

### DIFF
--- a/docker-compose-azure.yml
+++ b/docker-compose-azure.yml
@@ -1,19 +1,16 @@
 version: "3"
-
 services:
   app:
-    build: .
-    image: electic-io
+    image: noopkat/electric-io
     container_name: electric-io
     command: npm run start
     ports: # Can override if you have port collisions with something else
       - 3000:3000
-    env_file: # Expose to docker environment
-      - .env
     environment: # Expose to docker environment override
-      - SIMULATING
+      - SIMULATING=false
       - DEBUG
       - CONSUMER_GROUP
+      - CONNECTION_STRING=<IoTHubConnectionString>
     volumes: # For file editing, watching, etc
-      - ./:/usr/app
+      - /home/site/wwwroot:/usr/app
       - /usr/app/node_modules
--- a/docker-compose-azure.yml
+++ b/docker-compose-azure.yml
@@ -10,6 +10,7 @@ services:
       - SIMULATING=false
       - DEBUG
       - CONSUMER_GROUP
+      - PLATFORM=azure
       - CONNECTION_STRING=<IoTHubConnectionString>
     volumes: # For file editing, watching, etc
       - /home/site/wwwroot:/usr/app
--- a/docker-compose-azure.yml
+++ b/docker-compose-azure.yml
@@ -12,6 +12,8 @@ services:
       - CONSUMER_GROUP
       - PLATFORM=azure
       - CONNECTION_STRING=<IoTHubConnectionString>
+      - EDIT_MODE=unlocked
     volumes: # For file editing, watching, etc
-      - /home/site/wwwroot:/usr/app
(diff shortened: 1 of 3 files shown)


noopkat:
this looks good to me 👍

toolboc:
@mtheoryx ,  I am curious your thoughts on the approach to support cloud deployment over locally building from source in the docker-compose file.  For transparency, there is a very strong potential to have this work leveraged into a reusable workshop where participants deploy an instance of electric-io to Azure using the docker-compose.yml.

toolboc:
It looks like the fix to allow websockets on azure breaks websocket locally.  We may need to do something conditionally (i.e. presence of envrionment variable) in order to allow proper function in both places.

toolboc:
^ Husky lintter introduced in https://github.com/noopkat/electric-io/commit/f1a4b587ba6acc680eb82cb6de111bd9a377b10f seems to have formatted much of index.js

mtheoryx:
Awesome to see this @toolboc !

I am really glad to see that you split the compose file apart. I would defer to @noopkat but I feel most people would prefer to keep this as cloud-agnostic as possible. Compose for local development is really more of a convenience, but as you've demonstrated, can be useful for deployment.

AWS also supports deployments to ECS using docker compose: https://aws.amazon.com/about-aws/whats-new/2018/06/amazon-ecs-cli-supports-docker-compose-version-3/

I'm planning on proposing a change that would split the "server" from the "frontend" into 2 containers, and compose will be even more critical at that point for inter-container networking.

I'm very cool with this PR (even if the linter thing got wonky), and will let Suz merge it, or maybe even take a look at this on her show tomorrow!

docker-compose-azure.yml
```diff
@@ -0,0 +1,19 @@
+version: "3"
```
mtheoryx: You know, now that Im looking at it, I'm not sure there are even any real differences between this and the original compose file, other than the volume.

And technically, if you build just the container, volume mapping shouldn't be needed since we already copy everything required into the container. The volume mapping is only for if you want to persist to your host, or if you want to make changes on the host (developing) and have the reflected in the running container.

noopkat:
thanks @toolboc and @mtheoryx 

image

docker-compose-azure.yml
toolboc: @mtheoryx , as you mentioned, the main difference is that docker-compose-azure.yml will pull from the noopkat/electric-io repo, set the platform to Azure, and use a volume to persist data on the host.  This allows us a very easy mechanism to deploy to Azure and make modifications by reapplying the docker-compose file in the Azure Web App container settings.  It will also support our proposed workshop by allowing attendees to deploy, make changes, then lock their dashboard down and still have their previous changes persist.  

While the changes are subtle, the intent in each docker-compose file is different i.e. one supports deployment to Azure out of box, the other supports local dev.

PR #83, opened 2018-09-28
Make it more clear that we want contributions

mtheoryx:


noopkat:
Thanks, this is great!

image

PR #84, opened 2018-09-29
Update contributing list of contents

MichaelE1:
Hi! This PR relates to CONTRIBUTING.md

- Updated the list of contents
- Fixed a broken link in the list of contents
- Changed a reference from `electric-dreams` to `electric-io`

I also noticed that this document has a lot of overlap with README, particularly the installation part. Perhaps we could just link the reader back to README for this part? I haven't included that change in this PR, but it's something that might be worth discussing :mega:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -5,13 +5,17 @@ Hey thanks for your interest in contributing to this project! I really appreciat
 This page should get you up and running with how to run locally and how to adhere to this project's style guide also.
 
 1. [Installation](#installation)
-   1. [With a real Azure IoT Hub](#with-a-real-azure-iot-hub)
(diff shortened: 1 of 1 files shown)


mtheoryx:
Thank you for these additions!

On the topic of overlap, we may want to tweak both areas, actually. Here's what I'm thinking:

1. The Installation section in the Readme could be more about how to do a basic run with docker.
2. The installation part in Contributing could cover anything that would be additional for actually developing (thinks like test runners, volume mapping, etc)

The above is just my initial though. What are your thoughts?

mtheoryx:
> it would be a shame if anyone abandoned their attempts to try electric-io because they (erroneously) assumed that it could only be run with docker.

That's a really excellent point. Item 1 (installation in readme) should definitely include both approaches. I appreciate you taking the time to point that out.

PR #85, opened 2018-11-15
add azure web app compatibility

noopkat:
electric-io currently works fine inside a docker container on the azure web app platform, but needs some extra deployment files for vanilla azure web app deployments due to needing the dev dependencies to be pre-installed and the front-end to be built before everything boots.

I added the minimum required changes to get it up and running. The biggest major change is renaming the main server file from `index.js` to `server.js`. I hope this is ok. I think it's more descriptive anyway 😄 

Next steps in general are to create a production npm script but I figured it was slightly out of scope for this feature addition / bug fixes.

***update***: now with added tests / CI!

image


mpaarating:
Looks great! I don't have a way to actually test this, but I don't see any issues with the deploy file or anything.

noopkat:
@mpaarating this is an extremely good point. I am tempted to set up some devops to have as an integration test. Let's hold off merging and I can set this up.

test/azure-ci-test.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+
+node ./azure-puppeteer.js
```
noopkat: this might be the wrong relative path to this file, thoughts?

noopkat:
requesting @toolboc if they have a minute or two to look over these changes 🙇‍♀️

noopkat:
I also realise that anyone can bypass needing these changes by setting up an actual CI pipeline with Travis / Jenkins / Azure Devops / Circle / AWS Pipelines etc etc etc but this unlocks local git deploys specifically via Kudu running on the Azure Web App's instance.

noopkat:
it worked! 

image

PR #87, opened 2018-11-22
refactor server to use new event hubs package

### LINKED ISSUE #86: Update to use latest event hubs package
We need to deprecate the azure-iothub-receiver package and migrate the code to support @azure/event-hubs client library.

noopkat:
Fixed #86. Lots of work in here to do with clean up as well.

1. Use new event hubs package 🆕 
2. Run everything through Prettier where any previously committed files avoided reformatting 💅 
3. Abstract express routes into their own file 🎉 
4. Simplify simulating / not simulating logic in server file 😌 
5. Unify the three different objects exported from `liveHub` and `simHub` modules to export only one function: `startService`. This is the beginnings of work to create a plug-n-play interface for alternative platforms to Azure IoT Hub 🔌 

Nit: I dislike Prettier's 80 character line limit as it's mangling a fair amount of code within the routes file. However I read the Prettier docs and understand we're not meant to change it. Does anyone else agree that it made things _harder_ to read in some circumstances within this PR? 🙈


mtheoryx:
I think the 80 character limit is difficult on some things with Prettier, so I agree with you.

mtheoryx:
@noopkat If you want to add a custom `.prettierrc`, go for it.

Here's a sample from one of my projects: 
https://gist.github.com/mtheoryx/3a90ee30cf6bcbff65430b5a29faae1a

And here's a site where you can play around with the settings and decide what you prefer:
https://prettier.io/playground/

PR #91, opened 2019-01-28
sanitise all user inputs via express before saving data

noopkat:
Does what it says on the tin. 

I quickly tested this by spraying some simple `<script>alert("hacked")</script>` around the place, but a more formalised pen test would be cool if anyone knows of some auditing library we could use 🙇‍♀️ 

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,16 +18,18 @@
     "@azure/event-hubs": "^1.0.7",
     "azure-iothub": "1.8.2",
     "babel-runtime": "6.26.0",
+    "body-parser": "^1.18.3",
     "chart.js": "2.7.2",
     "dotenv": "5.0.0",
     "express": "4.16.0",
+    "express-sanitizer": "^1.0.4",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -18,16 +18,18 @@
...
     "express": "4.16.0",
+    "express-sanitizer": "^1.0.4",
     "marked": "0.4.0",
     "socket.io": "2.1.1",
     "vue": "2.5.16",
     "vue-focus": "2.1.0"
   },
   "engines": {
-    "node": ">= 6"
+    "node": ">= 10"
```
mtheoryx: I think, and others can correct me, that this is only effective if someone uses this as a library.

Regardless, it does make the node version support clear in the package.json

MichaelE1: Our readme also refers to the engines field for the version of node that should be installed to run the project. Assuming that docker isn't being used :sweat_smile:

PR #92, opened 2019-02-07
Update all dependencies (Babel 6 → 7)

kleinfreund:
Hey there,

I thought “What the hell, let’s update some dependencies” and there went my Thursday morning.

Big disclaimer: Although the few tests that exist pass and one can run the Webpack build, this might still break the hub. In particular, I don’t have a “Azure IoT Hub connection string”; thus, I can’t run the actual hub right now. The simulated hub can be started and seems to work, though! So, should things break, you can use this as a kind of draft for how updating dependencies might look like. :)

This pull request …

- upgrades all sorts of Babel things to Babel 7, and
- adds typescript as an explicit development dependency to workaround a bug in “vue-jest” (https://github.com/vuejs/vue-jest/issues/155). It can be removed from electric-io’s dev dependencies once the bug is fixed fixed.

Have a nice day. :robot: :fireworks:


noopkat:
Thanks so much for this tedious work @kleinfreund, you rule! Keeping electric-io healthy and up to date ✨ 

image

PR #93, opened 2019-02-17
Remove explicit TypeScript extension

kleinfreund:
Previously, electric-io had a explicit dependency on `typescript` which was only necessary due to https://github.com/vuejs/vue-jest/issues/155. Since that’s fixed, updating the version for `vue-jest` allows us to remove this dependency.


noopkat:
thanks @kleinfreund !
image

Issue #16, opened 2018-04-16
Support for sending payloads with device method calls


noopkat:

noopkat:
FYI I am currently working on this 🤚

noopkat:
wait, we did this already! 🎉 
closing.

PR #95, opened 2019-02-25
WIP: add support for device method payloads

noopkat:
Addresses #16 

When you call device methods via IoT Hub, you can also send payloads. I think it'd be great to support JSON payloads in electric-io when creating the Button cards to call device methods.

This is a draft PR. 

It looks like Prettier let loose on some files I have touched since we installed it.

TODO:
+ [ ] validate the JSON
+ [ ] add a test

image


benhalverson:
How does the user know that the payload should be a JSON object?

noopkat:
@benhalverson the label on the textarea form element advises the user to use valid JSON. JSON validator UX coming up soon.

PR #98, opened 2019-03-04
Refactor of card form templating

noopkat:
Upon hindsight, storing the card form template in the dashboard settings as template strings no longer needs to be necessary now that we can use regular Vue component files.

I refactored things to bring each individual card type's form fields back into Vue land.

I am extremely open to technical opinions on this - my Vue is very rusty, and I'm concerned that I am nesting these form fields perhaps one level deeper than necessary for things to make sense. Some fresh eyes would be much appreciated!

🙇‍♀️


mpaarating:
Looks **SO** much better!

PR #107, opened 2019-03-21
Added C2D message support on button cards ⚡️😍

bsatrom:
It would be cool to add cloud-to-device message support to the button card type for cases where you want to send a message to a device, but don't have a direct method to call, or don't need a response.

I'm not sure that this approach is 100% correct, but I added a drop-down to the button settings that lets you choose between calling a method or sending a message. 

Screenshot 2019-03-21 15 16 03

In addition, if you select a message, the method name input is hidden.

Screenshot 2019-03-21 15 15 46

Finally, since both types are sharing payload support, I made the name generic.

Let me know what you think! Thanks again for all your awesome work on electric-io!


noopkat:
hi @bsatrom,

This is absolutely fantastic - a really helpful new feature, and the user experience is spot on 💯 

Thank you!!

bitmoji

bsatrom:
Yay, you're welcome @noopkat and thank you again for such an awesome project!!! 🥇

PR #111, opened 2019-05-12
Add color picker to dashboard settings (an initial concept)

kleinfreund:
This pull request is a stepping stone towards #9 and adds a color picker to the dashboard settings field for the background color. It uses [vue-color](http://xiaokaike.github.io/vue-color).

Right off the bat, this implementation has severe shortcomings in terms of accessibility. Vue-color comes with a variety of picker styles, all of which only have the most basic keyboard accessibility (e.g. entering color names or values into a text input). For some of this components, one could implement more sophisticated versions; for example, navigating the compact color picker with arrow keys since it is laid out as a grid.

That said, this pull request is meant as an initial concept; hence, it only implements a color picker at one place in the UI, the dashboard settings. If it looks right, we can consider whether we want to continue integrating it, whether we want to use a different picker style, etc. I might be able to contribute more accessible components upstream, but I’d like to know first if the direction is good.

Ah, right. I also updated all Vue dependencies. Doing another round of dependency updates during the week after the active pull requests are merged. :)

See you in, well, actually less than half an hour! 🎇

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3577,6 +3577,11 @@
         "safe-buffer": "^5.0.1"
       }
     },
+    "clamp": {
+      "version": "1.0.1",
+      "resolved": "https://registry.npmjs.org/clamp/-/clamp-1.0.1.tgz",
+      "integrity": "sha1-ZqDmQBGBbjcZaCj9yMjBRzEshjQ="
+    },
     "class-utils": {
       "version": "0.3.6",
       "resolved": "https://registry.npmjs.org/class-utils/-/class-utils-0.3.6.tgz",
@@ -8172,6 +8177,11 @@
       "integrity": "sha1-7dFMgk4sycHgsKG0K7UhBRakJDg=",
       "dev": true
     },
+    "lodash.throttle": {
+      "version": "4.1.1",
+      "resolved": "https://registry.npmjs.org/lodash.throttle/-/lodash.throttle-4.1.1.tgz",
+      "integrity": "sha1-wj6RtxAkKscMN/HhzaknTMOb8vQ="
+    },
     "loose-envify": {
       "version": "1.3.1",
       "resolved": "https://registry.npmjs.org/loose-envify/-/loose-envify-1.3.1.tgz",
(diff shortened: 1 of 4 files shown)


escherina:
Just dropping a screenshot of the visual changes:
image

@kleinfreund Nice work! I think visually and functionally, this is fine, but I definitely think we could do better in terms of accessibility - as you pointed out, there's no keyboard access - although I do see markup in the `compact-picker` that indicates they have tried to make it accessible - but you just can't focus on any of the colour choices. Maybe we could contribute some changes to `vue-color` itself if the maintainers are open to it? (I would be happy to help out 😄)

Other than that, just one suggestion: I think it could be nice to have this picker hidden behind a collapsible section as it does take up a fair bit of space.

kleinfreund:
A note about improving accessibility of the compact color picker:

We should put the color picker in a disclosure widget (i.e. `<details>`). @escherina suggested making the individual color elements focusable so keyboard users could activate them. But the basic error here is the vue-color not using `button`s. I’d love to contribute that upstream. Or we build our own component. That would also be a great exercise to get me started with Vue. Hm. I gotta think about this some more.

escherina:
Some more notes, looking at the original `vue-color` package:

They're using the `role="listbox"` on a `ul`, each `li` has a `role` of `option` which does seem to match the [listbox spec](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/listbox_role) however I'm not sure why `listbox` is used given that it "is used for lists from which a user may select one or more items which are static". I think another structure with radio buttons (as we discussed on stream 😄) would be more appropriate.

kleinfreund:
There is a bug with this.

1. Choose a color with the picker and save.
2. Type in a color in the text field and save.

Expected:

New color is applied.

Actual:

New color is applied, but the old color is placed in the text field.

noopkat:
👨‍🍳 😘 👌 

This is great work. I'm merging this because it improves the experience for mouse input users out of the gate, without regressing the experience for keyboard users. Pushing improvements for a11y upstream would be such a great activity for everyone in this case. Feel free if anyone takes this on, to link this pull request to any upstream pull request.

Thanks also for fixing the 🐛 so fast as well ✨ 

bitmoji

PR #113, opened 2019-05-12
Update all the sweet dependencies

kleinfreund:
One more time with feeling. Flat out updating all dependencies to whatever is latest. Quick testing in simulated mode shows nothing wrong. Since there are major version updates to packages like dotenv and husky, it would probably be good to test this with a device. :)

Also, as always, feel free to merge off-stream. ✨


noopkat:
thank you again @kleinfreund 🙇‍♀ bitmoji

PR #115, opened 2019-05-19
Handle Errors in fetch calls to the api in configuration.js for getDashboard() and getDeviceList()

jhorner32:
Been following this project for a while via the Sunday stream so I wanted to take a look and see how I could contribute! 😄 

While I was browsing the code base, I noticed the configuration.js functions that are returning fetch calls aren't handling any errors. I believe there are a few more areas in the code base that also aren't handling any errors but since getDashboard() & getDeviceList() are essentially the initialization of rendering the dashboard/devices when you first load the application, I think this would be a good place to start handling errors.

### DIFF
--- a/.env
+++ b/.env
@@ -5,6 +5,6 @@
 
 # note: .env is a shell file so there can't be spaces around '=
 
-CONNECTION_STRING=''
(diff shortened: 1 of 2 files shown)


noopkat:
hey @jhorner32 this is a really nice and needed improvement to the experience of making these api calls from the front end 😄 

I particularly like how clearly laid out and well documented the nested ternary expressions are ✨ 

Thanks!

bitmoji

jhorner32:
Thanks for the feedback and reviewing this so quickly @noopkat! I'm looking forward to contributing more 😄

PR #117, opened 2019-05-21
Updated to DRY [kind of] code when handling errors in configuration.js

jhorner32:
Just some continued work in configuationl.js. This is leading off of the pr I sent on Sunday regarding handling errors for the api get/post requests. This v2 pr is to clean up/refactor some of the code from the first pr and create more reusable methods for any future api requests 😃I also refactored a tiny bit of the code to use es6 syntax.

Oh, and I noticed the es-lint that runs in the pre-commit hook does alter the format of the ternaries I'm using 😆I'm not sure if it's because of the code comments but that's likely the culprit.

Another quick note, I know fetch is a native javascript solution for fetching resources asynchronously across the network, but, it may be wise to adopt axios since it cuts out the middle step of passing the results of the http request to the .json() method (less code 🎉) . Axios just returns the data object you would expect, i.e:

axios
.then(response => console.log(response))

fetch
.then(response => response.json()).then(data => console.log(data))

Axios also tends to handle errors a bit more elegantly as well. I'd love to work on this refactor and set up if you think it's a good idea. Just a thought!!

Expect some prs to come from me over the next week or so to extend unit test code coverage on the vue components! 

*sorry for the novel lol*

### DIFF
--- a/public/js/lib/configuration.js
+++ b/public/js/lib/configuration.js
@@ -1,22 +1,7 @@
-export function getDashboard(vm, url) {
-  return new Promise((resolve, reject) =>
-    fetch((url = "/api/dashboard")).then(response => {
-      // Promise is resolved if we get a 200 response status code
-      response.ok
-        ? resolve(response.json())
-        : // If the response status ist't a 200, check if it's a 400
-        response.status === 404
-        ? // error response if the response status code is a 404
-          reject(`Request to ${url} failed with 404 Page Not Found`)
-        : // Catch all error response for any other response status code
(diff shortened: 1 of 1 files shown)


noopkat:
@jhorner32 thanks for tidying this up! It looks great and our config file also looks easier to read ✨ 

as for your axios suggestion - feel free to do it!

bitmoji

jhorner32:
You're welcome @noopkat! Sounds great, i'll start working on axios 😄✨

PR #119, opened 2019-06-02
Update unit test config/setup + AppComponent now has ~80% code coverage

jeffnhorner:
Hey Suz and noopcrew!

This pr surrounds minimal improvements to overall jest/unitTest config/setUp files as well as writing a unit test for the beastly (lolz) App Component that now has ~80% code coverage (an average):

Screen Shot 2019-06-02 at 8 29 09 AM

The App component has quite a bit going on lol so there's definitely room for improvement and increased code coverage here but I think this is a good start 😃 

Also, a quick note on the change I made here in the initial App data object regarding the SIMULATING variable: https://github.com/jhorner32/electric-io/blob/8920d44d7e03cbf6a4d48cbcea11092879dbc728/public/js/components/App.vue#L53

Jest was barking at me that SIMULATING (this is how it is used in master: https://github.com/noopkat/electric-io/blob/master/public/js/components/App.vue#L53) wasn't defined and when I more explicitly defined the "simulating" data object property value to its full env variable name of "process.env.SIMULATING", all was well. If you aren't a fan of this, I can do some additional research to see if there's another solution :)

Lastly, you probably noticed I separated the jest configuration from the package.json to it's own config file. I did this simply because it's a little cleaner and easier to view and identify the jest configuration.

Looking forward to your feedback! 🎉


noopkat:
Hi @jhorner32 !

Thank you SO MUCH for writing these tests!!!!!!!!!!!11111one
As for the `SIMULATING` global, I'm going to defer to @escherina on this one, because I believe they created the webpack global injection plugin to handle these globals IIRC 🤔 

We're super close to merging this, I just want a second opinion on that `SIMULATING` global reference.

bitmoji

jhorner32:
You're welcome!!!!!!1111one (lolz)

Thanks for the feedback. I actually did some additional research and you're right, it looks like the "SIMULATING" variable is defined globally in the webpack config using the DefinePlugin plugin: https://github.com/noopkat/electric-io/blob/master/webpack.config.js#L46-L48

So, good news, I just needed to add a global to the jest config: https://github.com/noopkat/electric-io/blob/0111eafe19280bc24a7dd9d7dbe24e81cfdb0c4b/jest.config.js#L25-L27

Quick fix!

Lastly, I reverted the changes I made to the App component data object here: https://github.com/noopkat/electric-io/blob/0111eafe19280bc24a7dd9d7dbe24e81cfdb0c4b/public/js/components/App.vue#L53

I'd definitely still say @kleinfreund @escherina should take a look, just to be sure, but I feel pretty good about this 😀

jeffnhorner:
Wanted to leave a quick comment about my username change! Any mentions to @jhorner32 will 404. My new username is @jeffnhorner.

bitmoji

public/js/components/specs/app.spec.js
```diff
@@ -0,0 +1,243 @@
...
+
+    expect(wrapper.vm.showSettings).toEqual(true);
+
+    wrapper.vm.dashboard.editMode = "locked";
+
+    expect(wrapper.vm.showSettings).toEqual(false);
+    expect(wrapper.find({ name: "dashboard-settings" }).exists()).toBe(false);
+  });
+
+  test("comput the appTitle with or without an emoji using the TITLE_EMOJI_REGEX constant", () => {
```
kleinfreund: Typo: “comput” should be “compute”.

jeffnhorner: Ah, good catch. Updated 😄

jest.config.js
```diff
@@ -0,0 +1,28 @@
...
+  collectCoverageFrom: [
+    "lib/**/*.js",
+    "public/js/**/*.{js,vue}",
+    "!public/js/dist/**",
+    "!**/node_modules/**"
+  ],
+  roots: ["lib", "public/js"],
+  setupFiles: ["./lib/testSetup.js"],
+  testURL: "http://localhost/",
+  husky: {
```
underscoredotspace: The `husky` option isn't part of the jest config and needs to remain in `package.json`. This is what's causing the Vaidation Warning when you run `npm t`.

underscoredotspace: The `husky` option isn't part of the jest config and needs to remain in `package.json`. This is what's causing the Validation Warning when you run `npm t`.

public/js/components/specs/app.spec.js
```diff
@@ -0,0 +1,243 @@
...
+  });
+
+  test("if child components mount", () => {
+    const wrapper = shallowMountApp();
+
+    expect(wrapper.find({ name: "dashboard-settings" }).exists()).toBe(true);
+    expect(wrapper.find({ name: "base-card" }).exists()).toBe(true);
+  });
+
+  test("check before component is created that the default data is clean", () => {
```
underscoredotspace: This doesn't seem to be a robust test. When I run `SIMULATING=true npm t` the environment variable is set correctly in node, and available as `process.env["SIMULATING"]`. In the test however, both the `SIMULATING` global and `vm.simulating` evaluate to `false`, hence the test is passing for the wrong reason. 

Unfortunately I don't have a suggestion on how to fix this, but maybe Jest/Vue testing legend @escherina does.

public/js/components/specs/app.spec.js
```diff
@@ -0,0 +1,243 @@
...
+        bgImageRepeat: true,
+        bgImageUrl: "",
+        title: "\u2700 IoT Dashboard"
+      };
+      expect(vm.saveDashboard).toHaveBeenCalled();
+    });
+
+    Vue.nextTick(() => {
+      expect(vm.dashboard).toEqual(
+        expect.objectContaining({
```
underscoredotspace: This is my new favourite Jest assertion 😍

underscoredotspace:
Oh, it may also be good to update `CONTRIBUTING.md` as line 111 refers to not having much coverage and NOW WE HAVE MORE THAN 30% 🎉🎉🎉🎉🎉

jest.config.js
jeffnhorner: Oh interesting, I was wondering why that warning was coming up. Looks like master currently has the husky config nested inside of the jest flag int he package.json so that's why I ported it over to the new jest.config.js file 😅  : https://github.com/noopkat/electric-io/blob/master/package.json#L72-L104

Updating 😄

public/js/components/specs/app.spec.js
jeffnhorner: Yeah reading back over this I would have to agree. I think in order for this to become more robust, I need to update the boolean value of SIMULATING and then re test the expected outcome.

i.e.

vm.SIMULATING = true;
expect(vm.simulating).toEqual(false);

vm.SIMULATING = true;
expect(vm.simulating).toEqual(true);

Something like that. ^ is untested so I could be completely wrong. I'll look into this when I get some down time today!

public/js/components/specs/app.spec.js
jeffnhorner: Right?! I love it too!

public/js/components/specs/app.spec.js
mpaarating: According to [this helpful issues](https://github.com/vuejs/vue-test-utils/issues/193#issuecomment-438437626), you should be able to just do `process.env = Object.assign(process.env, { SIMULATING: true });` in a `beforeAll()`. I haven't tested this myself.

jeffnhorner:
> Oh, it may also be good to update `CONTRIBUTING.md` as line 111 refers to not having much coverage and NOW WE HAVE MORE THAN 30% 🎉🎉🎉🎉🎉
> 
> Also, line 111 has smol typo 🤦‍♂ (welbeing)

Haha good idea, I'll definitely include that when I update the pr 😄

public/js/components/specs/app.spec.js
jeffnhorner: > According to [this helpful issues](https://github.com/vuejs/vue-test-utils/issues/193#issuecomment-438437626), you should be able to just do `process.env = Object.assign(process.env, { SIMULATING: true });` in a `beforeAll()`. I haven't tested this myself.

Ah, thanks for sending this over. I'll test this and take a deeper look when i update the pr. Thanks for the rec @mpaarating! 🎉

mpaarating: > > According to [this helpful issues](https://github.com/vuejs/vue-test-utils/issues/193#issuecomment-438437626), you should be able to just do `process.env = Object.assign(process.env, { SIMULATING: true });` in a `beforeAll()`. I haven't tested this myself.
> 
> Ah, thanks for sending this over. I'll test this and take a deeper look when i update the pr. Thanks for the rec @mpaarating! 🎉

So I've played around with this a bit and there's a few ways to solve this issue, but the underlying issue is that in `App.vue` `simulating` is being set to a string of `true` or `false` rather than a boolean.

You can either cast `SIMULATING` in `jest.config.js` to a string version of `true` or `false` based on `process.env.SIMULATING` or you can cast `simulating` in `App.vue` to a boolean based on the string value of `process.env.SIMULATING`. I personally prefer the cast to a string in the config rather than the messiness of keeping it in `App.vue`.

With that said, I have no idea why its treated as a boolean in one place and string in the other.

jest.config.js
```diff
@@ -0,0 +1,28 @@
...
+    hooks: {
+      "pre-commit": "pretty-quick --staged"
+    }
+  },
+  globals: {
+    SIMULATING: process.env.SIMULATING === true ? true : false
```
underscoredotspace: Maybe this will do it? And the same in `App.vue`
```suggestion
    SIMULATING: process.env.SIMULATING === "true" ? true : false
```

kleinfreund: Oh! That was actually a bug all along. The environment will always contain strings.

mpaarating: [unmask the villain](https://media.giphy.com/media/26DMUXQypGJaHhTsQ/giphy.gif)

jeffnhorner: So SIMULATING: process.env.SIMULATING === "true" ? true : false actually allows jest to evaluate if the app is running in SIMULATING mode or not.

@underscoredotspace @kleinfreund @mpaarating - my pushed commit (#4039a4d) shows this change a few others. You guys take a look when you get a minute and feel free to test that SIMULATING global variable in the App.spec.js test.

I would just change line 61 in App.spec.js to:

expect(vm.simulating).toEqual('test');

When you run SIMULATING=true npm t you will see the test will fail and show vm.simulating was expected to be true.

Then, run just npm t and you'll see see the test fail and show vm.simulating was expected to be false.

Thanks for all the help on this y'all! 😄

kleinfreund: This is looking good now.

- Running `npm test`: Both `vm.simulating` and `SIMULATING` evaluate to `false`
- Running `SIMULATING=true npm test`: Both `vm.simulating` and `SIMULATING` evaluate to `true`

You can add a `console.log` call to test this as well. No need to change the assertion to equal `test` to get the expected value. :)

public/js/components/specs/app.spec.js
kleinfreund: This is resolved with the recent commit.

kleinfreund:
:ship: it

jest.config.js
```diff
@@ -0,0 +1,23 @@
...
+    "lib/**/*.js",
+    "public/js/**/*.{js,vue}",
+    "!public/js/dist/**",
+    "!**/node_modules/**"
+  ],
+  roots: ["lib", "public/js"],
+  setupFiles: ["./lib/testSetup.js"],
+  testURL: "http://localhost/",
+  globals: {
+    SIMULATING: process.env.SIMULATING === "true" ? true : false
```
underscoredotspace: not a huge deal, but this would have the same effect with less words: 
```suggestion
    SIMULATING: process.env.SIMULATING === "true"
```

escherina:
Apologies for coming to this so late! I will see if I can squeeze a review in before the stream. Thank you very much @jeffnhorner for adding tests! 🎉🎉🎉

public/js/components/specs/app.spec.js
```diff
@@ -0,0 +1,243 @@
...
+        bgImageRepeat: expect.any(Boolean),
+        bgImageUrl: expect.any(String),
+        blockSize: expect.any(Array),
+        editMode: expect.any(String),
+        tiles: expect.any(Array),
+        title: expect.any(String)
+      })
+    );
+
+    vm.onSaveSettings(event => {
```
escherina: Another way to do this would be to access the `DashboardSettings.vue` file and mimic the event being emitted from the component. I find this is a bit more like the real behaviour the user will experience than calling the Vue method directly.

For example:

```javascript
import DashboardSettings from "../DashboardSettings.vue"
import * as configFns from "../../lib/configuration";

test("onSaveSettings method", () => {
  jest.spyOn(configFns, "saveDashboard");
  const wrapper = shallowMountApp();

  wrapper.find(DashboardSettings).vm.$emit("save-settings", {
    bgColor: "#fff",
    bgImageRepeat: true,
    bgImageUrl: "",
    title: "\u2700 IoT Dashboard"
  });

  expect(configFns.saveDashboard).toHaveBeenCalled();
})
```

This approach also fixes the issues with `vm.saveDashboard` I mentioned in my below comment. To check if `saveDashboard` was called with `.toHaveBeenCalled`, we have to spy on it. (Note: in this case, I imported it in a certain way so it was easier to spy on.)

public/js/components/specs/app.spec.js
```diff
@@ -0,0 +1,243 @@
...
+
+    expect(vm.appTitle).toEqual(
+      `<span class="hemoji">${title[1]}</span> ${title[2]}`
+    );
+  });
+
+  test("onSaveSettings method", () => {
+    const { vm } = shallowMountApp();
+
+    expect(vm.dashboard).toEqual(
```
escherina: This assertion fails for me when I run the test on its own (with `test.only`) - could someone do a quick sanity check and make sure that it's not just My Machine(tm)? 😄

public/js/components/specs/app.spec.js
```diff
@@ -0,0 +1,243 @@
...
+  test("onTileChange method", () => {
+    const { vm } = shallowMountApp();
+
+    expect(vm.dashboard).toEqual(mockDashboardData.dashboard);
+
+    mockDashboardData.dashboard.tiles[0].buttonText = "stop it";
+
+    vm.onTileChange(event => {
+      event.id = mockDashboardData.dashboard.tiles[0].id;
+      expect(vm.saveDashboard).toHaveBeenCalled();
```
escherina: This will fail - I believe the failure is masked by it being called inside the `vm.onTileChange` function. Notes and fix in the above comment.

public/js/components/specs/app.spec.js
```diff
@@ -0,0 +1,243 @@
...
+    expect(vm.dashboard.tiles.length).toBe(3);
+  });
+
+  test("onDeviceListRecieved method", () => {
+    const { vm } = shallowMountApp();
+    const io = jest.fn();
+    const socket = io();
+
+    vm.onDeviceListReceived(() => {
+      expect(socket).toHaveBeenCalled();
```
escherina: Couple of things here: the `onDeviceListReceived` is called inside the `created` hook of the Vue component, so we could get a bit closer to the actual behaviour of the component in the test by not calling `vm.onDeviceListReceived` directly, and instead mounting the component and asserting that `socket` was called.

Secondly, this `expect` is not reached by the test, as was the case with the two tests I commented on above. We should be able to run it like this, but it fails:

```javascript
vm.onDeviceListReceived();
expect(socket).toHaveBeenCalled();
```

I think it might be more useful to spy on the contents of the `socket.on` callback than to check if the `const socket = io();` ran correctly, We could do this with [this kind of approach](https://stackoverflow.com/questions/46204810/how-to-test-socket-io-events-being-called-with-sinon-spy) (it references `sinon.spy` but Jest can do something similar).

public/js/components/specs/app.spec.js
kleinfreund: Confirming that this test fails in isolation using the following command:

```sh
./node_modules/jest/bin/jest.js public/js/components/specs/app.spec.js -t 'onSaveSettings method'
```

underscoredotspace: yeah, fails for me. It's a type evaluation difference `"true"` vs `true`: 

```
"bgImageRepeat": "true", // test is bgImageRepeat: expect.any(Boolean),
```

underscoredotspace: Oh, but super weird that taking out `test.only` makes this change fail 🤔

public/js/components/specs/app.spec.js
```diff
@@ -0,0 +1,243 @@
+import { mount, shallowMount } from "@vue/test-utils";
```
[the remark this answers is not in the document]
jeffnhorner: I 100% agree! I'm a bit confused though, I actually already changed this file name to App.spec.js but my most recent commit doesn't show this. I'm staring at the file name right now in my local branch and it's App.spec.js but git says there's nothing to commit, working tree clean 🤔I'll figure it out lol.

public/js/components/specs/app.spec.js
```diff
@@ -0,0 +1,243 @@
...
+    vm.dashboard.bgImageRepeat = false;
+
+    expect(vm.dashStyle).toEqual({
+      backgroundColor: vm.dashboard.bgColor,
+      backgroundImage: vm.dashboard.bgImageUrl,
+      backgroundRepeat: "no-repeat"
+    });
+  });
+
+  test("h1 heading style based upon headingStyle computed method", () => {
```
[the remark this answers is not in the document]
jeffnhorner: Oh true! I really just wanted to make sure to try and cover as many lines and logic as possible 😄

public/js/components/specs/app.spec.js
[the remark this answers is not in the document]
kleinfreund: Rename the file to `app.spec.js` again and then run the following:

```sh
git mv public/js/components/specs/app.spec.js public/js/components/specs/App.spec.js
```

public/js/components/specs/app.spec.js
```diff
@@ -0,0 +1,243 @@
...
+    );
+
+    vm.onSaveSettings(event => {
+      event = {
+        bgColor: "#fff",
+        bgImageRepeat: true,
+        bgImageUrl: "",
+        title: "\u2700 IoT Dashboard"
+      };
+      expect(vm.saveDashboard).toHaveBeenCalled();
```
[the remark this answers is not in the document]
jeffnhorner: Good call, this is a really good point. I vaguely remember this would fail if it was outside of the vm.onSaveSettings (and on vm.onTileChange on line 162). I'll go ahead and remove this.

public/js/components/specs/app.spec.js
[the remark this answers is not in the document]
escherina: That makes sense! I think the code coverage approach is excellent for small utility libraries, but I've personally found that with Vue components, I get diminishing returns when trying to get higher code coverage. I find more bugs are caught by testing that the expected input to the component (e.g. props, emitted events, etc) yields the expected output. (Most of these tests are doing this already, so it's all good! 😊)

public/js/components/specs/app.spec.js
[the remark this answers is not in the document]
jeffnhorner: Thanks dude! @kleinfreund

public/js/components/specs/app.spec.js
jeffnhorner: Ah, I love this. A much more realistic behavior. I went ahead and updated both onSaveSettings and onTileCreate to adopt this type of test 😊

public/js/components/specs/app.spec.js
jeffnhorner: Good catch @escherina. Thanks for confirming @kleinfreund & @underscoredotspace.

This assertion should pass just fine with the changes I'm committing shortly.

public/js/components/specs/app.spec.js
jeffnhorner: Updated!

public/js/components/specs/app.spec.js
jeffnhorner: Ok gotcha - thanks for the detailed feedback here! Looking at the the mentioned approach now.

jeffnhorner: Ok, so, I've been having some difficulty writing a similar test that you mentioned [here](https://stackoverflow.com/questions/46204810/how-to-test-socket-io-events-being-called-with-sinon-spy). I updated the pr based upon all your feedback but left this as a TODO item. 

Just as reference, this is where I left off from the test: 
```
  test("onDeviceListReceived method", () => {
    const { vm } = shallowMountApp();
    const message = jest.fn();

    io().on("message", message);

    expect(message).toHaveBeenCalledWith(vm.onDeviceListReceived());
  });
```

and the error I was getting (jest mock function was never called): 
Screen Shot 2019-06-10 at 10 59 31 PM

I tried a few other variations but they all led to this error. Hopefully I'm missing something simple. I've never written a test for something like this before so it's a bit new to me 😅 (which is awesome).

But any who, I appreciate all the feedback and look forward to any additional feedback you have on this 😊

jest.config.js
jeffnhorner: Forgot to mention @underscoredotspace, thanks for this feedback! I updated this in my most recent commit.

public/js/components/specs/app.spec.js
escherina: Works perfectly now! 👍

public/js/components/specs/app.spec.js
escherina: To be honest, looking at this again, I'm not even sure I suggested the _right_ thing to test. 😄 We can assume that the `socket.io` library is thoroughly tested, so checking that `socket.on` was called probably isn't so useful for us.

What would be useful is to check that when `"message"` is received, `this.messages` is updated with the right information. So we could look into mocking socket.io events (maybe with a library [like `mock-socket`](https://github.com/thoov/mock-socket)) and checking the outcome, i.e. that `this.messages` is updated with the correct `message.body`. We could also check that the `if (this.messages.length > 500)` conditional behaves as expected.

However, I think this test could be a PR all on its own, so it shouldn't be a blocker for merging this good work. 👍

jeffnhorner: Ah, I see what you're saying. Well, I'll definitely be creating another pr soon and try and knock this one out 😅

jeffnhorner:
> All LGTM now. Thanks for making the requested changes - really nice PR! 😎🎉

Thanks so much for all of the feedback! #noopcrewFTW

noopkat:
thank you so much everyone for polishing this up! Again it's very exciting that we have such an improvement in  test coverage ✨ Merging now 🙇‍♀ 

bitmoji

PR #120, opened 2019-06-11
Prevent button hover from resizing parent container

underscoredotspace:
This recreates the `border` with `box-shadow`. `border` resizes the container when added which is visually unappealing, `box-shadow` does not. Visually, the button appears the same as when it had a border. 

Also removes inner focus ring in Firefox on `<input type="submit">` to be consistent with `<button>`

### DIFF
--- a/public/css/style.css
+++ b/public/css/style.css
@@ -182,6 +182,15 @@ input[type="submit"]::-moz-focus-inner {
 }
 
 .action-button {
+  --base-box-shadow: 1px 0 hsla(214, 45%, 30%, 1), 0 1px hsla(214, 45%, 40%, 1),
+    2px 1px hsla(214, 45%, 30%, 1), 1px 2px hsla(214, 45%, 40%, 1),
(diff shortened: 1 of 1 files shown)


underscoredotspace:
I had the same thought, but a pre-processor felt like a big step for a small amount of CSS. Custom Properties didn't occur to me, that's a very good idea as a short-term solution 👍. I'll update the PR for that. 

We could open an discussion issue for pre-processing/support perhaps?

kleinfreund:
@underscoredotspace I opened an issue for this already: https://github.com/noopkat/electric-io/issues/121

underscoredotspace:
This last commit using CSS custom properties will "break" IE, but I reckon you'd still get the functionality of the button states, just without the actual shadow. The fake border should work fine. I have no way of testing this just now. 

There's pretty good support for custom properties in all other modern browsers: https://caniuse.com/#search=custom%20properties

PR #124, opened 2019-06-17
Move cards with keyboard

kleinfreund:
This pull request adds keyboard navigation to cards (implements #4). It behaves like this:

- Cards can now be focused (i.e. they have `tabindex="0"`).
- If a card is focused *and* an arrow key is pressed, the card’s position will be updated in steps of 1 pixel (10 pixels if the `Shift` key is pressed) per keypress.

Additionally, I added a basic focus style to all elements and one to the cards in particular. This also improves focus styles for some unrelated elements (e.g. settings panel toggle, settings text inputs).

If I find the time this week, I will add the following things to this pull request:

- Prevent selecting text when dragging a card.
- Prevent cards from being dragged/moved off-screen.

**Note on accessibility**:

In order to make use of this feature, a user has to know about it. Currently, a user has no way of knowing what kind of interaction is possible when a card has focus. I’m not aware of an interactive element or accessible role that would convey “Hey, this element can be moved with arrow keys”. Therefore, there should be some kind of instruction explaining the possible interactions. I’m unsure how to go about this.

### DIFF
--- a/public/css/style.css
+++ b/public/css/style.css
@@ -502,3 +502,13 @@ dialog[open] {
   box-sizing: content-box;
   box-shadow: none;
 }
+
+/**
+ * Stops content from being selected (e.g. when moving around cards).
+ */
+body.dragging {
+  -webkit-user-select: none;
+  -moz-user-select: none;
+  -ms-user-select: none;
+  user-select: none;
+}
--- a/public/css/style.css
+++ b/public/css/style.css
@@ -8,6 +8,12 @@
   box-sizing: border-box;
 }
 
+:root {
+  --focus-color: hotpink;
+  --focus-outline: 2px solid var(--focus-color);
+  --focus-box-shadow: 0 0 0 2px var(--focus-color);
+}
+
 html,
 body {
   width: 100%;
@@ -18,6 +24,9 @@ body {
   position: relative;
   font-family: helvetica, arial, sans-serif;
 }
+
+:focus {
+  outline: var(--focus-outline);
 }
 
 .bold {
@@ -76,6 +85,11 @@ footer > a {
   border-radius: 6px;
   box-shadow: 0 0 15px rgba(0, 0, 0, 0.2);
   background-color: white;
+}
+
+.card:focus {
+  outline: none;
+  box-shadow: var(--focus-box-shadow);
 }
 
 .card:not(:last-child) {
--- a/public/css/style.css
+++ b/public/css/style.css
@@ -8,11 +8,16 @@
   box-sizing: border-box;
 }
 
+html,
+body {
+  width: 100%;
+  height: 100%;
+}
+
 body {
   position: relative;
   font-family: helvetica, arial, sans-serif;
-  width: 100%;
-  height: 100%;
+}
 }
 
 .bold {
@@ -99,7 +104,7 @@ h1 {
   color: black;
   font-family: "Nanum Pen Script", sans-serif;
   font-size: 50px;
-  margin-top: 0px;
+  margin-top: 0;
 }
 
 h1 .hemoji {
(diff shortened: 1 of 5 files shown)


noopkat:
Hi @kleinfreund - I absolutely love where this is going 🎉 

I think the things you'd like to add soon are good ideas 💡 

I love the pink focus style for the focusable elements on the page and think we should keep it for almost all of the elements with one exception - the focus outline around the cards themselves. The pink will either be highly visible or not depending on the background colour chosen for the dashboard itself. I was thinking perhaps we could aim for something a bit more tolerant of this? 

I did some [CSS crimes in a Codepen](https://codepen.io/noopkat-the-bashful/pen/ydJbEZ) earlier today to explore what a compact, almost 100% CSS solution of an auto-contrasting focus state might look like. The only downside to my method is that Edge browser doesn't currently support CSS filters, but does support CSS variables. So on Edge 15 onwards, the card focus style will be set to the background colour custom property but not inverted 😭  So I'm happy to shelve this experiment for a little while in order to play around a bit further.

Again thanks for tacking this challenging issue; you thought of many details and it's amazing to see.

noopkat:
P.S. letting the user know they can move the card is an interesting challenge as you mentioned. We could do a couple of things:

1. Show a little help bubble the first time they focus on a card (we could make this screenreader friendly too) and then not show them again. Video game tutorial style.
2. Permanently show a concise message along the top of the card when focused
3. Additionally we could also include a link to a "shortcut help" modal in the settings card too as an extra prompt

Having the dashboard in locked mode will hide the edit and delete card buttons, so if we went with solution 2 above we could also hide that consistent help UI in locked mode also.

underscoredotspace:
One solution to the background contrast issue, would be to have the outline colour be pink or yellow (?) depending on the contrast - we already have a function that can tell us if the contrast is sufficient. The `--focus-color`variable could be set when the background is. 

Also @noopkat, I'd prefer to call your Pen "civil disobedience" 😉

kleinfreund:
Content will no longer be selectable while dragging around cards with the next commit.

I also reconsidered the idea to prevent cards from being moved off-screen. I think this needs some more thought. Currently, cards can be moved off-screen with the mouse. The dashboard’s viewport even adapts to this and expands its size. Should we limit this behavior to one axis? It would still be possible to position cards "below the fold", but not to the left or right. Maybe the current behavior is fine for the time being.

Regarding focus styles: I think we should evaluate that in a separate issue/pull request. I didn’t pay much attention to staying in line with one consistent style (e.g. the buttons use a completely different way to visually represent focus). We can find a more consistent visual design for focus and the dashboard in general (getting rid of that pesky cyan background in the settings panel for example :P)

kleinfreund:
@noopkat CSS filters should work with the Edge versions we’re targetting: https://caniuse.com/#feat=css-filters.

noopkat:
@kleinfreund @underscoredotspace thank you both for weighing in! Also - my incorrect Edge support comment came about because I misread the caniuse.com recommendation initially and thought that Edge supported the `url()` function _only_ - instead of the other way around 🤦‍♂ 

As for moving cards out of bounds - let's push this discussion to another issue so we don't increase scope too much on this feature :shipit:

public/js/components/BaseCard.vue
```diff
@@ -113,6 +113,16 @@ export default {
     };
   },
 
+  watch: {
+    draggingWithMouse: function(dragging) {
+      if (dragging) {
+        document.body.classList.add("dragging");
```
noopkat: this is much more elegant than what I implemented in the past to prevent selection. Thank you 🙏🏻

public/css/style.css
```diff
@@ -66,13 +80,35 @@ footer > a {
...
   box-shadow: 0 0 15px rgba(0, 0, 0, 0.2);
   background-color: white;
 }
 
+.card:focus {
+  outline: none;
+}
+
+.card:focus::before {
+  --focus-offset: calc(-2 * var(--card-border-width));
```
noopkat: 🤓

noopkat:
this turned out so great thank you so much @kleinfreund ✨ 

bitmoji

Issue #125, opened 2019-06-18
Discuss: Should it be possible to position cards off-screen? If yes, how?


kleinfreund:
In its current form, electric-io’s cards can be moved off-screen with the mouse. With #124 merged, that is true for keyboard, too.

Is this exptected behavior?

There are more or less three options:

- Allow cards to be moved off-screen (current behavior)
- Don’t allow cards to be moved off-screen
- Allow cards to be moved off-screen on the vertical axis only

I bring this up because one could make the point that the dashboard gets rather hard to access if we continue to allow cards to be positioned wherever. The Web Content Accessibility Guidelines have a success criterion called [Reflow](https://www.w3.org/TR/2018/REC-WCAG21-20180605/#reflow) that deals with this question in the context of a web page with a more typical layout (i.e. boxes are stacked on the block axis, creating a vertically-overflowing page).

underscoredotspace:
My web-based [portfolio](https://messy.cloud) uses windows that are logically similar to our card concept. There I have chosen to not allow windows to leave the viewport. This does however increase the logic required in the move and size functions, and we'd need to handle viewport resizing/zooming. 

In my app there is a minimum height and width for a window. When the viewport is resized the window's size is reduced until it reaches the minimum, then the position is moved up and left until it reaches (0,0). Using the viewport size as a trigger also handles zoom-in which gives consistent (and logical) behaviour. [Testing this out has caused me to notice some bugs, so thanks for triggering this thought 😆]

Perhaps the easiest route though (from a "logic required" viewpoint) would be to allow cards to be moved anywhere and provide scroll bars. This option would have the added benefit of making the app more usable on small screens too. 

Despite the wall of text I've just put in here, I don't have a strong opinion. Just thought it would be of value to share my experience.

kleinfreund:
@underscoredotspace Your portfolio also provides a clue for why free positioning of cards should be possible: All OSes I know allow windows to be moved out of their viewport freely. Atleast, when using the mouse to drag windows around.

noopkat:
@underscoredotspace this is off topic but your portfolio site is very cool!

noopkat:
I like @mtheoryx's idea of "bouncing" the card back if dragged or keyboard moved off the top or left edges of the screen. I think right and bottom edges should just extend the dashboard itself. Thoughts?

mjasnowski:
[...]

PR #127, opened 2019-06-26
Refine card movement behavior

kleinfreund:
I did some work on the overall behavior of card movement. This is partially related to #125.

- Card movement via mouse or keyboard is now capped off at the top-left corner. It simply stops there for the momement. There is no bouncy-back business. And no, I also did not add the Pokémon you’ve-hit-a-wall sound, but I think that it is an incredibly good idea to do so.
- Refactored the dragging code to avoid re-registering the `mousemove` event listener every time the card is clicked.
- Fixed a bug with keyboard navigation not emitting its new tile position which lead to the card position not being saved if it was moved the the keyboard last.
- All background styles of the dashboard are applied to the body element now to ensure that they extend to all available space on the page, not just the dimensions of the dashboard. For this, I needed to drop the related `dashStyle` test since this computed property no longer exists. I haven’t added adequate tests instead because I don’t know how (yet). See https://github.com/noopkat/electric-io/pull/127/commits/ba6bf412d26d6364d21a370eddbcacd4f2c8c2d7.
- Cards can no longer be moved while in editing mode. Allowing card movement in editing mode resulted in loss of data since the card component will be re-rendered when its data updates. Since a card’s position is part of its data, the card will be re-rendered with the currently stored data; thus, it dropped any data entered in the form fields.
- Cards can now be dragged with both mouse-based *and* touch-based pointer devices. See https://github.com/noopkat/electric-io/pull/127/commits/a01d262e157a1283e8d7da6f1bc0fc6b835f3030.
- A card’s editing form now has a cancel button that moves the card back into its viewing mode without saving any changes.

I had another idea:

While dragging, one could hold down `Shift` to snap to the card to a 10 pixel grid. What do you think about that?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7616,7 +7616,8 @@
     "js-tokens": {
       "version": "3.0.2",
       "resolved": "https://registry.npmjs.org/js-tokens/-/js-tokens-3.0.2.tgz",
-      "integrity": "sha1-mGbfOVECEw449/mWvOtlRDIJwls="
+      "integrity": "sha1-mGbfOVECEw449/mWvOtlRDIJwls=",
+      "dev": true
     },
     "js-yaml": {
       "version": "3.13.1",
@@ -7921,6 +7922,7 @@
       "version": "1.3.1",
       "resolved": "https://registry.npmjs.org/loose-envify/-/loose-envify-1.3.1.tgz",
       "integrity": "sha1-0aitM/qc4OcT1l/dCsi3SNR4yEg=",
+      "dev": true,
       "requires": {
         "js-tokens": "^3.0.0"
       }
@@ -11624,14 +11626,6 @@
         "lodash.throttle": "^4.0.0",
         "material-colors": "^1.0.0",
         "tinycolor2": "^1.1.2"
-      }
-    },
-    "vue-focus": {
-      "version": "2.1.0",
-      "resolved": "https://registry.npmjs.org/vue-focus/-/vue-focus-2.1.0.tgz",
-      "integrity": "sha1-egM3zpB01e8D0VpLW4Ys9F5eBOM=",
-      "requires": {
-        "loose-envify": "^1.2.0"
       }
     },
     "vue-hot-reload-api": {
--- a/package.json
+++ b/package.json
@@ -29,7 +29,6 @@
     "vue": "^2.6.10",
     "vue-a11y-dialog": "^0.5.0",
     "vue-color": "^2.7.0",
-    "vue-focus": "^2.1.0",
     "whatwg-fetch": "^3.0.0"
   },
   "engines": {
--- a/public/css/style.css
+++ b/public/css/style.css
@@ -107,6 +107,7 @@ footer > a {
   border: var(--card-border-width) dotted var(--background-color);
   border-radius: var(--card-border-radius);
   filter: invert(100%);
+  pointer-events: none;
 }
 
 .card:not(:last-child) {
--- a/public/css/style.css
+++ b/public/css/style.css
@@ -94,6 +94,15 @@ footer > a {
   background-color: white;
 }
 
+.card--is-being-dragged:hover {
+  cursor: move;
+  user-select: none;
+}
+
+.card:not(.card--is-being-edited) {
+  cursor: grab;
+}
+
 .card:focus {
   outline: none;
 }
@@ -260,11 +269,6 @@ input[type="submit"]::-moz-focus-inner {
 .action-button:active {
   transform: translate(4px, 4px);
   box-shadow: none;
-}
-
-.card.dragging:hover {
-  cursor: move;
-  user-select: none;
 }
 
 .card form input[type="submit"] {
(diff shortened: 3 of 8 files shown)


kleinfreund:
I want to give a reason for why the interface for updating a card’s position looks like below:

```js
updateCardPosition({ x = this.x, y = this.y }) {
  // …
}
```

This method can be called like this:

```js
updateCardPosition({ x: 20, y: 10 });
updateCardPosition({ x: 20 });
updateCardPosition({ y: 10 });
```

This means that both parameters, x and y, are optional. You can pass any object and be sure that the variables `x` and `y` are always set to something useful (i.e. the current position of the card). Only if the passed object has properties x or y will these variables be set with something else.

Using default parameters wouldn’t work here because only the last parameter in an argument list can be a default parameter.

Off topic: Ugh, the text box jumping around while typing is so annoying.

noopkat:
> And no, I also did not add the Pokémon you’ve-hit-a-wall sound, but I think that it is an incredibly good idea to do so.

😄 🎉

kleinfreund:
Prior to the latest two commits, this pull request made editing a card impossible. That’s now fixed.

underscoredotspace:
Are you still working on this? I've spotted a couple of minor things, both are when a card is at the past the bottom or right of the initial viewport:

- When the card is dragged back into the initial viewport the scroll bar adjusts, which causes the card and mouse pointer to lose sync. 
- When you scroll the window there is no background. 

I like the change in general, it feels very intuitive. Good work! 😃

With regards to your shift + drag question, I think this is a very good idea but we'd need to think about how we signal that this is available.

kleinfreund:
@underscoredotspace These bugs are present in either version of electric-io and should be solved separately, I think. I’ll look into them before tomorrow’s stream to see which of these are related to this pull request.

kleinfreund:
The card’s title input field has a `v-focus` attribute that causes the element to steal focus when a card is in editing mode and an arrow key is pressed. Therefore, a card in editing mode cannot be moved more than once with arrow keys. I’ll see if I can work around this.

underscoredotspace:
Personally, I find focus stealing to be very frustrating. Do we need that behavior at all here?

kleinfreund:
With the latest commits:

- All background styles are applied to the body element now to ensure that they extend to all available space on the page, not just the dimensions of the dashboard. For this, I needed to drop the related `dashStyle` test since this computed property no longer exists. I haven’t added adequate tests instead because I don’t know how (yet).
- Moving a card in editing mode with the keyboard now works more than once since the input field no longer keeps stealing focus.

kleinfreund:
With https://github.com/noopkat/electric-io/pull/127/commits/a01d262e157a1283e8d7da6f1bc0fc6b835f3030, cards can be dragged with all pointers (i.e. mouse-based and touch-based pointers).

kleinfreund:
Finally, https://github.com/noopkat/electric-io/pull/127/commits/f043fe09ff501fda14fe826fe097c01fa26c02b2 fixes #130. Moving a card via the keyboard no longer emits the position changes to the parent component immediately. Instead, this is done once when the `beforeunload` event is fired (see [MDN onbeforeunload](https://developer.mozilla.org/en-US/docs/Web/API/WindowEventHandlers/onbeforeunload)). I don’t have experience with this particular event type so I don’t quite no if that’s 100% reliable. That said, I don’t know why it shouldn’t work.

underscoredotspace:
I'm still seeing the issue from #130 in your latest commit on Firefox. The card is also refusing to move with mouse event, but moves with keyboard in edit mode. 

On mobile, touch dragging is working nicely for me, but I notice that at one card is always stuck to the bottom of the viewport. 

Sorry I probably don't have time to do a more thorough review today, from what I've seen I'd be inclined to approve this and raise bugs for what remains unless anything else jumps out. I've given you a few drive-by comments which hopefully is helpful rather than irritating! 😬 

I'm happy that you've achieved what you set out to with this PR :shipit:

kleinfreund:
I jumped the gun with my approach of using the `beforeunload` event for making sure that the card position is saved. Since an asynchronous request stores the dashboard, triggering the saving process just before unloading a document *might* sometimes store the dashboard successfully, but more likely than not, it will cancel the request as soon as the synchronous part of the `beforeunload` event listener is executed.

I’m looking into this again to see about a better approach for saving a card’s position while editing.

@underscoredotspace I can no longer reproduce any data loss on card movement. Weird. 🤔

kleinfreund:
I think I’m done here now. This pull request is already way too bug and a bad example of how one should structure a contribution: Keep it small. I piggy-backed a lot of things into one pull request here. For the sake of not blowing it up any further, I stopped worrying about making cards editable and movable at the same time and just disabled that: When a card is being edited, it may not be moved. Otherwise, data would be lost with the current data flow architecture.

Phew. :sweat_smile:

noopkat:
Thank you @kleinfreund and @underscoredotspace for spending so much time on this - the outcome is very thoughtful and thorough 💯 

Any bugs that are the result of existing issues before this feature was rolled out should definitely be addressed in new issues and pull requests.

bitmoji

PR #128, opened 2019-06-28
Nested payload properties

tkoster:
Hi @noopkat and friends :wave:,

I recently discovered your stream through a friend who is just starting out to code on Twitch. Seeing you and your community work together to build and learn is inspiring :star::gem:.

To give something back, I wanted to contribute a pull request that adds support for nested message payload properties and address issue #8 :gift:.

I usually don't tune in live because of time zones :australia::sleeping: but I'll see if I can work something out if this PR looks like it's getting close to a merge :coffee::coffee::coffee::coffee:.

-- Thomas

---

## Approach

I have replaced the existing code in `LineChartCard.vue` and `NumberCard.vue` that extracts a property value from the message according to the card's "Data Property" with a function that uses JMESPath.

### JMESPath

The [jmespath.js](https://github.com/jmespath/jmespath.js) library is a JavaScript implementation of [JMESPath](http://jmespath.org/): a way to select a (possibly nested) value from a JavaScript object or array. This is a new dependency added in this PR.

Although JMESPath is more powerful than we need for issue #8, it is conveniently backwards compatible with the current way cards use the "Data Property". Maybe the library's advanced features might come in handy when working with devices that send inconveniently complex messages.

Still, jmespath.js is quite modest compared to alternatives and has no dependencies. Webpack reports that it increases the bundle size by about 21 KiB (when built with `--production`).

jmespath.js uses the [Apache-2.0 license](https://www.apache.org/licenses/LICENSE-2.0.txt).

### Errors

An invalid path would silently cause no values to appear on the card, similar to what happens today if the property is missing from a message. In future, we can look at preventing invalid paths by checking that the "Data Property" is a valid JMESPath in the card settings UI.

As before, if you select a value from the message that is not a number, the card won't display it.

### Performance

The line chart re-evaluates all messages (up to the 500 limit) every tick and discards all but the last 20 that are truthy. This is a bit wasteful but I didn't want to refactor this unless we know for sure it's a problem.

#### Existing dashboards

This change should be almost completely backwards compatible - I believe most existing dashboard configurations should work with no modifications. The exception would be dashboards that use message properties whose names contain characters that are [special to JMESPath](http://jmespath.org/specification.html). These are mostly non-alphanumeric characters that would be ridiculous to use in property names anyway.

#### MQTT

This change is all about JavaScript objects and JSON but I have no idea if/how this might affect the work in progress on MQTT. I'm actually not all that familiar with IoT :sweat_smile:.

## Testing

I have included tests for the new function.

However, it didn't seem right to add nested properties to the simulation mode as they appear to be modelled on real devices, so I left the simulation mode alone. Let me know if you'd like me to add some nested objects to the simulation mode.

Otherwise, you can test using Azure IoT Hub and the simulated device in the [Azure IoT Hub Quickstart](https://github.com/Azure-Samples/azure-iot-samples-node). Change [this line in the simulated device](https://github.com/Azure-Samples/azure-iot-samples-node/blob/c3abae6b766338db8530d727abe1ecbf94aab696/iot-hub/Quickstarts/simulated-device/SimulatedDevice.js#L32-L35) to send messages with nested objects or arrays. This is how I recorded the demo below.

## Examples

- `temperature` the "temperature" property
- `samples[0]` the 0th value in the "samples" array
- `measurement.humidity` the "humidity" property of the "measurement" object

See the [JMESPath tutorial](http://jmespath.org/tutorial.html) and [JMESPath examples](http://jmespath.org/examples.html) for more.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7607,6 +7607,11 @@
         }
       }
     },
+    "jmespath": {
+      "version": "0.15.0",
+      "resolved": "https://registry.npmjs.org/jmespath/-/jmespath-0.15.0.tgz",
+      "integrity": "sha1-o/Iiqarp+Wb10nx5ZRDigJF2Qhc="
+    },
     "js-levenshtein": {
       "version": "1.1.6",
       "resolved": "https://registry.npmjs.org/js-levenshtein/-/js-levenshtein-1.1.6.tgz",
--- a/package.json
+++ b/package.json
@@ -24,6 +24,7 @@
     "dotenv": "^8.0.0",
     "express": "^4.16.4",
     "express-sanitizer": "^1.0.5",
+    "jmespath": "^0.15.0",
     "marked": "^0.6.2",
     "socket.io": "^2.2.0",
     "vue": "^2.6.10",
--- a/public/js/components/LineChartCard.vue
+++ b/public/js/components/LineChartCard.vue
@@ -5,6 +5,7 @@
 <script>
 import Chart from "chart.js";
 import chartOptions from "../lib/chartOptions.js";
+import { evaluatePath } from "../lib/messagePropertyEvaluation.js";
 
 export default {
   name: "line-chart-card",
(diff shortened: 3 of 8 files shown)


noopkat:
Hi @tkoster !!

This pull request is so good 💯 🏅 Thanks for all of the attention to detail in both the code and also the pull request notes ✨ The GIF is superb and the license / file size / example details are also so helpful.

Welcome to the Electric-IO contributor family! ❤️ 

bitmoji

PR #129, opened 2019-07-05
Additional Code Coverage

jeffnhorner:
Hey Suz and noopcrew!

Here's a pr (nothing crazy or super exciting) for some additional unit test coverage for the frontend components in electric io's code base.

The coverage report is lighting up with a lot more green now (yay!): 

Screen Shot 2019-07-05 at 7 51 01 AM

These tests weren't nearly extensive as the main App component that I extended coverage for a few weeks back but it may still be good to have some more eyes on this (@kleinfreund, @underscoredotspace, @escherina ?). Just a thought 😊 

Looking forward to any feedback and getting this pushed through!

P.S. I'm working on another pr to finish up the code coverage for the rest of the frontend components 🎉


public/js/components/specs/NumberCard.spec.js
```diff
@@ -1,24 +1,61 @@
 import { shallowMount } from "@vue/test-utils";
 import NumberCard from "../NumberCard";
 
-describe("Number card", () => {
+describe("NumberCard", () => {
```
tkoster: Does the snapshot for `NumberCard` need to be updated to reflect this rename?

[`public/js/components/specs/__snapshots__/NumberCard.spec.js.snap`](https://github.com/noopkat/electric-io/blob/0737221d33778c09142f3cbfd0736012fe7e2126/public/js/components/specs/__snapshots__/NumberCard.spec.js.snap#L3)

kleinfreund: Yep, I think so. @jeffnhorner Run the following to update the snapshot accordingly:

```sh
npm test -- -u
```

kleinfreund:
@jeffnhorner The `whatwg-fetch` development dependency is missing; hence, the tests fail on your branch with a fresh clone. Run the following to add it:

```sh
npm install --save-dev whatwg-fetch
```

With the change suggested in @tkoster’s comment, the tests should pass in CI, too.

jeffnhorner:
Thanks for the feedback @tkoster & @kleinfreund - updates were pushed in my most recent commit. All checks are now passing ☺️

noopkat:
Hi @jeffnhorner these new tests are fantastic and this PR is exciting despite your protests that it is not 😄 

Thank you for all the hard work on bringing this code base up to great coverage so far. I think it's my turn to start writing unit tests for the backend code now! 💪 🙇‍♀ 

bitmoji

jeffnhorner:
Hi @noopkat! Thanks for reviewing and merging this in 😄 My apologies on not making a note on the pre-commit but it looks like you figured out what I was doing.

Full disclosure, I think unit tests are very exciting! However, I tend to devalue the work behind writing test 😅. Thanks for the affirmation though, I need to stop thinking this way lol.

Oh, and I was finally able to watch through the stream! I'm actually in CST so only an hour behind, but, 11am is usually when my wife and I are at church so I'm usually unable to catch the live stream. We actually may start attending the early service so I look forward to being in the live stream soon 🎉 

Long winded ^ lolz but thanks again and happy contributing/coding!

PR #132, opened 2019-07-07
Axe core unit tests

knitcodemonkey:
This commit adds `axe-core` to all the jest tests. 

By default, axe runs the entire suite of axe tests. You can define the rules, or ignore rules, by adding an `config` parameter a la [this jest example](https://github.com/dequelabs/axe-core/blob/develop/doc/examples/jest_react/link.test.js)

At the moment, all tests pass.

Screen Shot 2019-07-07 at 11 52 18 AM


noopkat:
Oh my gosh I am so excited that we now have a11y testing as a merge blocker!!!!!!!!!!
Thank you so much @knitcodemonkey 🎉 ❤️ 

And thanks to all contributors who helped fix enough a11y issues in order for this PR to be possible ❤️ 

bitmoji

PR #135, opened 2019-07-08
Added Short Hex Color Code parsing to RGB in Color Contraster

tusharsrivastava:
In `colorContraster.js` I've added the logic to process short HEX color code. I've used `String.prototype.repeat` but I doubt we need to add some polyfill for IE support for that. 

Please advice.

### DIFF
--- a/public/js/lib/colorContraster.js
+++ b/public/js/lib/colorContraster.js
@@ -83,27 +83,16 @@ function validHexColor(inputHexString) {
   // check if string is either 4 or 7 characters long and is string
   // added 9 for #RGBA types value (we remove alpha part later using substring)
   // Also, RGBA don't have short code
-  if (
-    ![4, 7, 9].includes(inputHexString.length) ||
-    typeof inputHexString !== "string"
-  ) {
+  // Edit: Removed string check as it doesn't make sense here
+  if (![4, 7, 9].includes(inputHexString.length)) {
     return false;
   }
 
   // Another small check to validate valid HEX input
-  // 1. Remove '#' from string
-  // 2. Split into array of chars
-  // 3. Map over each entry to check if parseInt with base 16 yields valid number
-  // 4. Finally test if a NaN is present.
-  // If NaN were present, this condition will be true implying invalid Hex code
-  // It might be an overkill here
-  if (
-    Array.from(inputHexString.substr(1)).find(char =>
-      isNaN(parseInt(char, 16))
-    ) !== undefined
-  ) {
-    return false;
-  }
+  // A simple Regular Expression can handle the validation here
 
-  return true;
+  // 1 or more characters between 0-f both cases (captial and small)
+  const exp = /^[a-fA-F0-9]+$/;
+
+  return exp.test(inputHexString.substr(1));
 }
--- a/public/js/lib/colorContraster.js
+++ b/public/js/lib/colorContraster.js
@@ -31,7 +31,7 @@ function parseColor(color) {
 
 function parseHex(color) {
   // #ff00de
-  const r; const g; const b, const justColor;
+  let r; let g; let b; let justColor;
   
   // need to check here if a 3 digit short hex code is provided
   // We're assuming here that the color will always be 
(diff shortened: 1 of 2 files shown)


tusharsrivastava:
@underscoredotspace Thanks for approval. Further since I'm new to this repo; thanks to @noopkat live stream that I got introduced to it just yesterday; I'm seeing that we don't have enough error handling on `parseHex` function. For now, I've not altered it as I'm assuming there's a contract somewhere that guarantee correct input to this function. Otherwise, we need to implement that as well. Right?

underscoredotspace:
Yeah @tusharsrivastava, I think that would be a good addition. Maybe return `null` if it doesn't validate? Then we could provide an error response in the component that uses the function. 

Do you want to add to this PR for the Hex function then you can raise an issue for the other colour functions? Don't want this kind of PR to get too large.

public/js/lib/colorContraster.js
```diff
@@ -31,12 +31,24 @@ function parseColor(color) {
 
 function parseHex(color) {
   // #ff00de
-  // support for 3 character hex strings would be nice :)
-  // pull request me if you like - noopkat
-  const justColor = color.substring(1, 7);
-  const r = parseInt(justColor.substring(0, 2), 16); // hexToR
-  const g = parseInt(justColor.substring(2, 4), 16); // hexToG
-  const b = parseInt(justColor.substring(4, 6), 16); // hexToB
+  let r; let g; let b; let justColor;
```
underscoredotspace: Since you're making additions, this can be written out more simply:

```suggestion
  let r,  g, b,  justColor;
```

public/js/lib/colorContraster.js
```diff
@@ -31,12 +31,24 @@ function parseColor(color) {
...
+  
+  // need to check here if a 3 digit short hex code is provided
+  // We're assuming here that the color will always be 
+  // (simple string length test would do)
+  if (color.length === 4) {
+    justColor = color.substring(1, 4);
+    // Using [String.prototype.repeat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/repeat)
```
underscoredotspace: I don't think we need this comment since we're supporting es6+ browsers now.

tusharsrivastava: @underscoredotspace I believe you're right. Let me remove the comment as well

public/js/lib/colorContraster.js
```diff
@@ -31,12 +31,44 @@ function parseColor(color) {
...
+  }
+  // Another small check to validate valid HEX input
+  // 1. Remove '#' from string
+  // 2. Split into array of chars
+  // 3. Map over each entry to check if parseInt with base 16 yields valid number
+  // 4. Finally test if a NaN is present.
+  // If NaN were present, this condition will be true implying invalid Hex code
+  // It might be an overkill here
+  if (Array.from(color.substr(1))
+       .map(b => isNaN(parseInt(b, 16)))
```
underscoredotspace: This works, but could be done in 1 step with `.find(char => isNaN(parseInt(char, 16)))`. Given how large this function is now, I'd extract the validation to an `validHexColor()` function.

tusharsrivastava: Sure Let me update that

underscoredotspace:
@tusharsrivastava I like your last commit, but I understand how people feel about RegEx so I'd like to hear from some other collaborators before we merge this. Otherwise we're good to go. 

@noopkat @kleinfreund Any thoughts?

tusharsrivastava:
I understand this, that's why I added a comment above my expression to explain what it's doing as well

kleinfreund:
I’m fine with short regular expressions. `^[a-fA-F0-9]+$` is okay to read in my opinion.

underscoredotspace:
Thanks @kleinfreund, just keeping myself in check. I'm happy with the RegEx and everything works as expected. Thanks for your contribution @tusharsrivastava!

tusharsrivastava:
I'm happy I could help

noopkat:
This is a super neat fix thanks @tusharsrivastava! ✨ 

Thanks for reviewing, @underscoredotspace and @kleinfreund 🙇‍♀ 

bitmoji

PR #136, opened 2019-07-12
Validate data property input on card settings

tkoster:
Hi @noopkat and team,

I was so pleased with the positive reactions to my previous contribution :joy: that the temptation to submit another bougie pull request was just too great! :tada:

I also felt bad for increasing the net number of `TODO` comments in electric-io with #128 :disappointed:

---

This pull request adds validation to the settings form for the number card and line chart card so that users know when their data property is not a valid JMESPath.

## New component

To reduce duplication in the view, there is a new component for the data property field shared by both the number card and line chart card settings forms.

The component validates the data property input and displays a message beside the field if it is invalid.

There is a CSS variable for the colour - nice fire engine red - but it clashes a bit with the hot pink focus outline. You can also see that the card grows taller when the error message is visible, which I'm not a fan of. Suggestions for improving the visual design are very welcome! **Roast my Vue!** :sweat_smile:

## Accessibility

The `aria-invalid` and `aria-describedby` attributes are set on the data property field when the value is invalid.

eio-not-a-jmespath-aria-2

Note also that an invalid data property value does not prevent the card from being saved. I offer you two excuses for this: :grin:

1. I don't know Vue.js very well and couldn't figure out how to make it work :dizzy_face: !

2. I believe it's a more human-friendly experience to the let users experiment safely and save their work, even if it's incomplete. Since a mistake is reversible and an invalid input only causes a harmless empty/zero card, Electric-IO won't break and the user won't damage their dashboard permanently. Form validation on the web is notorious for poor design and putting up walls in front of the user does them no favours here. However, feedback on the _front_ of the card that something is wrong in the settings would be an improvement to think about later.

I sadly do not encounter many opportunities to work on accessible software in my day job (because "enterprise" :weary:) so I'm a bit out of my depth about what to do for users using accessibility aids. So please, **roast my a11y too!** :sweat_smile:

## Changes to messagePathEvaluation.js

I have added a `pathIsValid` function to `messagePathEvaluation.js` for components to use.

However, while extending `messagePathEvaluation.js` to support validation, I encountered one bug of my own making and two bugs in jmespath.js:

1. It was only catching `ParserError`, but jmespath.js also throws `LexerError` and `RuntimeError` if things go awry. The function now accounts for all three of these errors.

2. The parser in jmespath.js throws `Parsererror` (sic) in one of its branches ([source code](https://github.com/jmespath/jmespath.js/blob/b16deb4f32270b2c75c08460e6afd2b064547a2a/jmespath.js#L764)), hence my use of `toLowerCase` to catch these.

3. The parser in jmespath.js also fails to reject a path ending in a `.` which causes an unseemly `TypeError` in the console later. This is [already reported](https://github.com/jmespath/jmespath.js/issues/36) upstream but not fixed. I worked around this.

Overall, the apparent lack of attention to the issues and open PRs on jmespath.js makes me a little bit nervous about it - I first heard of JMESPath while working with the Azure CLI (Python) - but I haven't encountered anything yet in jmespath.js we can't easily avoid. The library still does _a lot_ for us.

## Testing

The component is tested so that
- its `name` prop is used for the `input` element, and
- the `aria-invalid` and `aria-describedby` attributes are set correctly

However, moving the `label` and `input` elements for the data property into a separate component broke one of the tests contributed by @jeffnhorner: "that there are 3 labels in the LineChartSettings window". This was because the unit tests use a shallow mount.

I have replaced this test with three new tests that hopefully give us similar confidence: to check that the fields required to save line chart card settings are present on the form. The tests look for the `name` attribute (which the new component also has).

The tests for `pathIsValid` are mostly there to ensure somebody simplifying `messagePathEvaluation.js` doesn't cause the jmespath.js bug to bite us later.

### DIFF
--- a/public/js/components/DataPropertyField.vue
+++ b/public/js/components/DataPropertyField.vue
@@ -1,17 +1,20 @@
 <template>
-  <label v-bind:for="inputId" v-bind:class="{ invalid: !isValid }">
+  <label v-bind:for="inputElementId" v-bind:class="{ invalid: !isValid }">
     Data Property (supports
     <a href="http://jmespath.org/tutorial.html" target="_blank" rel="noopener"
       >JMESPath</a
     >)
     <input
       type="text"
-      v-bind:id="inputId"
+      v-bind:id="inputElementId"
       v-bind:name="name"
       v-model="model"
       v-bind:aria-invalid="!isValid"
+      v-bind:aria-describedby="!isValid ? errorElementId : null"
     />
-    <span v-if="!isValid">This is not a valid JMESPath.</span>
+    <span v-bind:id="errorElementId" v-if="!isValid"
+      >This is not a valid JMESPath.</span
+    >
   </label>
 </template>
 
@@ -33,8 +36,11 @@ export default {
     isValid: function() {
       return pathIsValid(this.model);
     },
-    inputId: function() {
+    inputElementId: function() {
       return `data-property-input-${this.tileId}`;
+    },
+    errorElementId: function() {
+      return `data-property-error-${this.tileId}`;
     }
   }
 };
--- a/public/js/components/DataPropertyField.vue
+++ b/public/js/components/DataPropertyField.vue
@@ -1,21 +1,28 @@
 <template>
   <label v-bind:class="{ invalid: !isValid }">
     Data Property (supports
-    <a href="http://jmespath.org/tutorial.html" target="_blank">JMESPath</a>)
-    <input type="text" v-bind:name="name" v-model="model" v-bind:aria-invalid="!isValid" />
+    <a href="http://jmespath.org/tutorial.html" target="_blank" rel="noopener"
+      >JMESPath</a
+    >)
+    <input
+      type="text"
+      v-bind:name="name"
+      v-model="model"
+      v-bind:aria-invalid="!isValid"
+    />
     <span v-if="!isValid">This is not a valid JMESPath.</span>
   </label>
 </template>
 
 <script>
-import { pathIsValid } from '../lib/messagePropertyEvaluation.js';
+import { pathIsValid } from "../lib/messagePropertyEvaluation.js";
 
 export default {
   name: "data-property-field",
   props: ["name", "value"],
   data() {
     return {
-      model: "",
+      model: ""
     };
   },
   mounted() {
--- a/public/js/components/DataPropertyField.vue
+++ b/public/js/components/DataPropertyField.vue
@@ -1,11 +1,12 @@
 <template>
-  <label v-bind:class="{ invalid: !isValid }">
+  <label v-bind:for="inputId" v-bind:class="{ invalid: !isValid }">
     Data Property (supports
     <a href="http://jmespath.org/tutorial.html" target="_blank" rel="noopener"
       >JMESPath</a
     >)
     <input
       type="text"
+      v-bind:id="inputId"
       v-bind:name="name"
       v-model="model"
       v-bind:aria-invalid="!isValid"
@@ -19,7 +20,7 @@ import { pathIsValid } from "../lib/messagePropertyEvaluation.js";
 
 export default {
   name: "data-property-field",
-  props: ["name", "value"],
+  props: ["tileId", "name", "value"],
   data() {
     return {
       model: ""
@@ -31,6 +32,9 @@ export default {
   computed: {
     isValid: function() {
       return pathIsValid(this.model);
+    },
+    inputId: function() {
+      return `data-property-input-${this.tileId}`;
     }
   }
 };
--- a/public/js/components/DataPropertyField.vue
+++ b/public/js/components/DataPropertyField.vue
@@ -1,0 +1,30 @@
+<template>
+  <label v-bind:class="{ invalid: !isValid }">
+    Data Property (supports
+    <a href="http://jmespath.org/tutorial.html" target="_blank">JMESPath</a>)
+    <input type="text" v-bind:name="name" v-model="model" v-bind:aria-invalid="!isValid" />
+    <span v-if="!isValid">This is not a valid JMESPath.</span>
+  </label>
+</template>
+
+<script>
+import { pathIsValid } from '../lib/messagePropertyEvaluation.js';
+
+export default {
+  name: "data-property-field",
+  props: ["name", "value"],
+  data() {
+    return {
+      model: "",
+    };
+  },
+  mounted() {
+    this.model = this.value;
+  },
+  computed: {
+    isValid: function() {
+      return pathIsValid(this.model);
+    }
+  }
+};
+</script>
--- a/public/js/components/NumberCard.vue
+++ b/public/js/components/NumberCard.vue
@@ -28,7 +28,7 @@ export default {
       const lastMessage = this.messages.pop();
       if (lastMessage) {
         const value = evaluatePath(this.tile.property, lastMessage);
-        this.number = value === null ? "" : parseFloat(value).toFixed(1);
+        this.number = parseFloat(value).toFixed(1);
       } else {
         this.number = 0;
       }
--- a/public/js/components/NumberCard.vue
+++ b/public/js/components/NumberCard.vue
@@ -28,7 +28,7 @@ export default {
       const lastMessage = this.messages.pop();
       if (lastMessage) {
         const value = evaluatePath(this.tile.property, lastMessage);
-        this.number = parseFloat(value).toFixed(1);
+        this.number = value ? parseFloat(value).toFixed(1) : "";
       } else {
         this.number = 0;
       }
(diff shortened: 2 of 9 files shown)


public/js/components/DataPropertyField.vue
```diff
@@ -0,0 +1,30 @@
+<template>
+  <label v-bind:class="{ invalid: !isValid }">
+    Data Property (supports
+    <a href="http://jmespath.org/tutorial.html" target="_blank">JMESPath</a>)
+    <input type="text" v-bind:name="name" v-model="model" v-bind:aria-invalid="!isValid" />
```
kleinfreund: The label should also be explicitly associated with the input element via a `for` attribute on the label and a matching `id` attribute on the input. Explicity label-control association is better supported by screen readers than implicit association (i.e. nesting the input inside the label).

public/js/components/DataPropertyField.vue
```diff
@@ -0,0 +1,30 @@
+<template>
+  <label v-bind:class="{ invalid: !isValid }">
+    Data Property (supports
+    <a href="http://jmespath.org/tutorial.html" target="_blank">JMESPath</a>)
```
kleinfreund: In _my_ opinion, links should almost never have `target="_blank"`. It should be the user’s decision whether a link is opened in a new window/tab or not. Adding this attribute takes this decision away from the user and I don’t see a good reason for that in this use case.

I would like to hear some opinions on this. :)

(I realize that this was present in the code before.)

public/js/components/NumberCard.vue
```diff
@@ -28,7 +28,7 @@ export default {
       const lastMessage = this.messages.pop();
       if (lastMessage) {
         const value = evaluatePath(this.tile.property, lastMessage);
-        this.number = parseFloat(value).toFixed(1);
+        this.number = value === null ? "" : parseFloat(value).toFixed(1);
```
kleinfreund: I don’t think that’s particularly helpful. With `NaN`, the user can assume that something’s wrong with the data. An empty string hides this. Now it looks more like a glitch in the UI. Sometimes there are values and sometimes there aren’t.

public/js/lib/messagePropertyEvaluation.js
```diff
@@ -10,18 +10,45 @@ import * as jmespath from "jmespath";
...
+    const value = jmespath.search(input, path);
+    return { value: value, isValidPath: true };
   } catch (error) {
-    if (error.name === "ParserError") {
-      // TODO: The user entered a path that is not a valid JMESPath. This should be detected and prevented earlier in the card settings UI.
-      // Return no match.
-      return null;
+    switch ((error.name || "").toLowerCase()) {
+      case "lexererror":
+      case "parsererror":
```
kleinfreund: If we don’t know what a `ParserError` or `LexerError` means and we don’t do anything with them, they should not appear in the code at all.

public/js/components/DataPropertyField.vue
tkoster: OK. How do you recommend I generate a unique ID to use for this? Everything the internet suggests looks like workarounds to me.

public/js/components/DataPropertyField.vue
escherina: I personally don't mind having the link open in a new window here, but if we use `target="_blank"` then it should be accompanied by `rel="noopener"` for security reasons (explanation here: https://developers.google.com/web/tools/lighthouse/audits/noopener)

tkoster: I agree. I followed the precedent set by the "what's this?" link on the sticker card settings for the alt text. However, navigating away in this case will lose the user's unsaved changes.

public/js/components/DataPropertyField.vue
escherina: In the last stream Suz generated a unique ID using the `tile.id`: https://github.com/noopkat/electric-io/blob/master/public/js/components/BaseCard.vue#L54

In the `DataPropertyField.vue` component, we could accept the `tile.id` as a prop `tileId`, and then use it in the input ID (e.g. `<input :id="'data-property-input-${tileId}'">`

kleinfreund: This is a very common problem in frameworks like Vue.js. One way I see recommended a lot for handling this in Vue is creating a [global mixin](https://vuejs.org/v2/guide/mixins.html#Global-Mixin) like suggested in this comment: https://github.com/vuejs/vue/issues/5886#issuecomment-308647738. Another option is using a third part package like [vue-uniq-ids](https://www.npmjs.com/package/vue-uniq-ids).

This might be a bit out of scope for this pull request, so feel free to ignore this. This is not a blocker for merging this pull request.

public/js/components/NumberCard.vue
tkoster: Fair enough. Reverted.

public/js/components/DataPropertyField.vue
```diff
@@ -0,0 +1,30 @@
...
+    <input type="text" v-bind:name="name" v-model="model" v-bind:aria-invalid="!isValid" />
+    <span v-if="!isValid">This is not a valid JMESPath.</span>
```
escherina: Extra accessibility suggestion: We could give this element an ID as well (using the same unique ID technique as suggested above) and then add something like:
```
:aria-describedby="!isValid ? `data-property-error-${tileId}` : ''"
```
to the input element. This would associate the error message with the input when the input is invalid.

Not a blocker for merging, just a nice-to-have! 😄

public/js/lib/messagePropertyEvaluation.js
tkoster: Although jmespath.js does not document these in detail, a quick scan of its code reveals that:

- it throws `LexerError` if it fails to tokenize the path, e.g. if the user entered a completely unrecognizable character like `#`
- it throws `ParserError` if it fails to parse the token stream, e.g. if the user entered an incomplete path like `array[`
- it throws `RuntimeError` if its evaluating a bad slice, e.g. `array[0:0:0]`

In all these cases, I think our response is the same: return `null` to the card and report the path is invalid, hence all three fall through to the same `return`.

Leaving the cases out causes these errors to end up in the console. Also I think Vue binding doesn't do so well with an unhandled exception from `pathIsValid`.

kleinfreund: If the response is the same and we don’t actually handle the cases differently, I suggest to reflect that in the code as well:

```js
return { value: null, isValidPath: false };
throw error;
```

Since the switch case statement doesn’t contain any `break`s, this statement does the same if the error name is one of the three.

public/js/components/DataPropertyField.vue
tkoster: Great suggestion, I'll try it. Should this `span` move outside the `label` too?

public/js/lib/messagePropertyEvaluation.js
tkoster: Sorry, I don't quite follow.

> Since the switch case statement doesn’t contain any breaks, this statement does the same if the error name is one of the three.

Yes. Are you saying that's wrong?

public/js/components/DataPropertyField.vue
kleinfreund: This is an interesting scenario. For one thing, messages informing a user about invalid input should be part of the label because that causes them to be read out when a screen reader is faced with the related control. However, as @escherina suggested, we already create a relation between the validation message and the form control. I’m not quite sure if we’re over-complicating matters with this or not.

In short, I’d definitely keep the message inside the label to ensure it’s announced along with the primary label content.

public/js/lib/messagePropertyEvaluation.js
kleinfreund: In short, switch statements without `break`s are quite weird.

Consider this:

```js
function f(num) {
  switch (num) {
    case 1:
    case 2:
      return 2;
    default:
      return NaN;
  }
}

f(2) // returns 2
f(1) // returns 2
```

Calling the function `f` with either `1` or `2` as an argument will return `2` in both cases! I find that very surprising. When I read this code for `num = 1`, I think it will evaluate the first case statement as true and do nothing since it is empty, but for some reason, the second case statement’s block will be executed as well **without evaluating `case 2`** at all.

To me, that is incredibly unexpected. In our case, I would not expect the code to return that object at all for `ParserError` or `LexerError`. But it does. And that’s very confusing.

Since you want to return the same for all three errors, just do that and drop the switch case statement altogether. It doesn’t provide anything is just very confusing to read.

tkoster: Oh I see your point now. My background includes many curly brace languages where `switch` is used for this specific effect and I did not consider that some contributors might find this behaviour surprising. Banning fall-through cases is not uncommon in style guides - for example Google's C++ and Java style guides ban it in most situations.

I'll change this to an `if`.

noopkat:
Hi @tkoster,

Thank you so much for circling back on this feature and adding validation to the field. This PR write-up is also very 💯 !

I like the approach of allowing 'incremental editing / saving' as I relate to this on other webapps when I'm wanting to just save other changes with the incomplete work.

Thanks @kleinfreund and @escherina for the reviews 🙇‍♀ 

bitmoji

PR #137, opened 2019-07-14
Accessible color picker

kleinfreund:
I’ve been working on a more accessible color picker implementation over the last two months und this is what I got. Feel free to not merge this. This is more of a nice-to-have and might be completely overkill. Also feel free to improve upon this. The implementation in Vue is a port of a pure JavaScript implementation and uses custom reactivity instead of relying on Vue. This is just the way it was implemented. We could make use of more Vue as well. ¯\_(ツ)_/¯

First things first: It turns out that depending on what they do, color pickers can be quite the big beasts. Previously, I added [vue-color](https://github.com/xiaokaike/vue-color) ([on bundlephobia](https://bundlephobia.com/result?p=vue-color@2.7.0)) to the dashboard settings in order to kick off work for #9. With this pull request, there is now a custom color picker implementation plus it is used in three places: dashboard settings, line chart cards, and number cards.

## Bundle size

Before:

```sh
👻 ll public/js/dist
-rw-r--r-- 1 phil phil  77K Jul 14 15:05 main.bundle.js
-rw-r--r-- 1 phil phil 737K Jul 14 15:05 vendors~main.bundle.js
```

After:

```sh
👻 ll public/js/dist
-rw-r--r-- 1 phil phil 101K Jul 14 15:56 main.bundle.js
-rw-r--r-- 1 phil phil 660K Jul 14 15:56 vendors~main.bundle.js
```

Looking at this, the main bundle size has increased dramatically due to pulling in the color picker in two additional places. This is somewhat expected since the current version only uses the vue-color picker in one place.

However, the vendors bundle size has decreased by an even larger amount.

## Picture

Screenshot from 2019-07-14 16-08-04

## Moving picture

peek9EBX4Z

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,6 @@
     "socket.io": "^2.2.0",
     "vue": "^2.6.10",
     "vue-a11y-dialog": "^0.5.0",
-    "vue-color": "^2.7.0",
     "whatwg-fetch": "^3.0.0"
   },
   "engines": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -3631,11 +3631,6 @@
         "safe-buffer": "^5.0.1"
       }
     },
-    "clamp": {
-      "version": "1.0.1",
-      "resolved": "https://registry.npmjs.org/clamp/-/clamp-1.0.1.tgz",
-      "integrity": "sha1-ZqDmQBGBbjcZaCj9yMjBRzEshjQ="
-    },
     "class-utils": {
       "version": "0.3.6",
       "resolved": "https://registry.npmjs.org/class-utils/-/class-utils-0.3.6.tgz",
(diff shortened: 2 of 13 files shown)


kleinfreund:
Oh I forgot to mention the most exciting thing about this and why it even deserves to have the term “accessibility” in the title:

Keyboard navigation! I have left out some things like navigating the range inputs in steps of 10. Can add them later. First, I need to hit the super market. :) See y’all later. <3

Also, please merge #136 first. There will be conflicts. :)

noopkat:
This is such an amazing improvement on the existing color picker ✨ 
I can see just how much work went into this, and how many hours of tweaking it must have taken to get everything working so smoothly! 💯 

Thank you so much for this contribution.

bitmoji

package.json
```diff
@@ -29,7 +29,6 @@
     "socket.io": "^2.2.0",
     "vue": "^2.6.10",
     "vue-a11y-dialog": "^0.5.0",
-    "vue-color": "^2.7.0",
```
noopkat: 🎉

PR #139, opened 2019-07-14
Refine card sizes

kleinfreund:
This is a quick follow-up to #137 where I changed the card sizing behavior so that the editing forms had enough room for their content. This had some unwanted side effects like the sticker card being as big as its contained image. This is now fixed.

Cards now only grow in size when editing them.

### DIFF
--- a/public/css/style.css
+++ b/public/css/style.css
@@ -96,6 +96,7 @@ footer > a {
   --card-border-radius: 6px;
   --card-border-width: 2px;
 
+  width: var(--card-tile-width);
   position: relative;
   padding: 15px;
   border: var(--card-border-width) solid #222;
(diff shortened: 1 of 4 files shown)


public/css/style.css
```diff
@@ -113,6 +114,11 @@ footer > a {
   cursor: grab;
```
noopkat: _I love this_

noopkat:
I hunted through the bitmoji registry for a chef kiss and it exists! So I'm using that. 👩‍🍳 👌 
Your PR is just in time for me to be spying on Lucy the Calathea plant again, and her camera stream took over the entire electric-io dashboard until I tested out this PR. Thanks for going back and polishing up the regression 💯 

bitmoji

PR #142, opened 2019-07-19
fix bug on editingCard variable setting when editing a card (loses changes)

noopkat:
I noticed that there was a slight regression with handling dragging event filtering for when a card is in the editing state. If you click anywhere on the card after changing an input when in the card settings, it will reset the changes you made. This could not be reproduced when using keyboard tabbing to navigate and also save the card. So I thought it might have something with do with the app knowing if the card is being edited or not when a mousedown event is fired.

Here is a GIF showing the behaviour:

captured

I did a quick manual test and it seems to have fixed the problem. It might be worth adding a test for this - thoughts? I'm about to be offline for a week so feel free to merge this if it looks good, and we can add a test later I promise 🙇

### DIFF
--- a/public/js/components/BaseCard.vue
+++ b/public/js/components/BaseCard.vue
@@ -16,7 +16,7 @@
         <button
           class="inline-button edit-button"
           ref="editButton"
-          @click="editingCard = true"
+          @click="() => editingCard = true"
         >
           edit
         </button>


noopkat:
update: it appears we may have to install something like enzyme in order to write a robust test for this. I think @knitcodemonkey was suggesting we use this a little while ago 👍

kleinfreund:
Good morning Suz.

I looked into this and I cannot reproduce this bug.

Additonally, the change you propose doesn’t alter the way the `editingCard` state is changed. As far as I know, Vue will execute the content within `@click=""` as an expression. It doesn’t have to be an explicit function. That is, both `editingCard = true` and `() => editingCard = true` produce the same outcome.

Could you go back and verify whether you can verify the bug on the `master` branch? I get the feeling that maybe you tried this before the build process was fully done. I certainly remember this being a bug that I looked into when working on the cards.

Have a nice weekend. :)

noopkat:
@kleinfreund I just checked this on master and can no longer reproduce 🤔 
I _swear_ that I waited for the build to complete, and ensured that the server was running in the background. I'm not sure why this was happening in the end, but it could have been because I had an old version of master that I branched off without realising? 🤷‍♀ 

Thank you for looking into this, and for explaining that Vue can just run code without it being an explicit function - TIL! I still very much feel like a Vue noob so thank you for your patience as well 🙇‍♀ 

Closing this!

PR #143, opened 2019-07-20
Maintenance!

kleinfreund:
Hallo!

YAMPR — Yet Another Maintenance Pull Request

What do we have today?

- The usual: Updated dependencies. Alas, jest and friends net us 14953 outstanding security vulnerabilites. The actual offender, of course, is lodash.
- The annoying: I added an `.eslintrc.json` because I was annoyed by Vue files not being linted. To get some actual linting going in VS Code, I also needed to add `eslint` and `eslint-plugin-vue` to electric-io. I also needed to disable a couple of Vue-specific linting rules since they clashed directly with `pretty-quick` — after fixing a linter warning, pretty-quick would re-introduce the faulty code. Running in circles. We’re still obeying to pretty-quick. _Still_ getting used to it.
- The chore: I went through all .vue files and fixed all linter errors and warnings. In general, our .vue files now have more consistent code styles (e.g. opting for the `:` shorthand instead of `v-bind:`, etc.).

Have a beautiful Saturday everyone. 👋


kleinfreund:
I want to look into whether it’s really necessary to bunlde `eslint` and `eslint-plugin-vue` with electric-io in order to get linting working for Vue. Maybe it’s sufficient to install these globally for the developers who want to actually use them.

kleinfreund:
So I think we do want to have our own bundled copies of `eslint` and `eslint-plugin-vue` so that we’re able to read in our own code style preferences via the `.eslintrc.json` file. There are a couple of things that need to be configured there so that they don’t conflict with `pretty-quick`.

I also added types to all Vue props and fixed tests that were breaking due to this.

noopkat:
thank you @kleinfreund !

PR #146, opened 2019-07-21
Clean up tests a bit

kleinfreund:
- Removed `whatwg-fetch` from dependencies (only needed as devDependency)
- <del>Removed all Axe tests.</del><ins>Skipped all Axe tests.</ins> We never actually tested anything with them. The Axe tests just passed because we were testing some kind of empty component/page fragment. Introducing a deliberate error like an unlabeled input element or an empty heading didn’t break the tests. I currently see no way of getting these tests to work properly. :(
- Added `@types/jest` as a devDependency. This removes the errors on test-related functions like `describe` and `test` that are shown in some editors/IDEs (e.g. VS Code)
- Streamlined the `shallowMountComponent` helper function across the tests so it is used the same way in all tests. This should make it easier for future contributors to understand what’s going on there.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1872,6 +1872,21 @@
         "@types/istanbul-lib-report": "*"
       }
     },
+    "@types/jest": {
+      "version": "24.0.15",
+      "resolved": "https://registry.npmjs.org/@types/jest/-/jest-24.0.15.tgz",
+      "integrity": "sha512-MU1HIvWUme74stAoc3mgAi+aMlgKOudgEvQDIm1v4RkrDudBh1T+NFp5sftpBAdXdx1J0PbdpJ+M2EsSOi1djA==",
+      "dev": true,
+      "requires": {
+        "@types/jest-diff": "*"
+      }
+    },
+    "@types/jest-diff": {
+      "version": "20.0.1",
+      "resolved": "https://registry.npmjs.org/@types/jest-diff/-/jest-diff-20.0.1.tgz",
+      "integrity": "sha512-yALhelO3i0hqZwhjtcr6dYyaLoCHbAMshwtj6cGxTvHZAKXHsYGdff6E8EPw3xLKY0ELUTQ69Q1rQiJENnccMA==",
+      "dev": true
+    },
     "@types/node": {
       "version": "12.6.8",
       "resolved": "https://registry.npmjs.org/@types/node/-/node-12.6.8.tgz",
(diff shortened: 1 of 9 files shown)


kleinfreund:
> If I recall correctly, adding `jest` to the list of environments in the eslint config file will remove the issues with the global `test` and `describe` keywords.

Unfortunately, for me, this didn’t change the linter errors. Googling after this doesn’t yield many helpful results.

> I think it would be great if we could look into fixing the Axe tests (in a separate PR, of course) rather than remove the setup that @knitcodemonkey put in.

Indeed, that would be awesome. But I coudln’t figure out how to make it work. I tried a series of things to get Axe to run on actual content, but always came up empty. When I find the time, I’ll look into this again.

escherina:
Yes, let's leave the dependency in, as this is likely happening to others too. (Out of interest, is it the eslint linter, or a TS error? I'm wondering what I did with my extension setup 😄)

No problem if you don't have time to fix it - you've done a lot of PRs lately! But I think it would be easier for someone else to fix the bug if the axe-related code was left in the repo - would you mind adding it back in? (Maybe @knitcodemonkey might have time to take another look at it, or could point us in the right direction?) I'll open an issue now so we have visibility on getting it fixed.

kleinfreund:
@escherina I added the Axe tests again, but disabled them and left a comment.

The error regarding the test suite methods is indeed a error from the TypeScript language server:

```
Cannot find name 'describe'. Do you need to install type definitions for a test runner? Try `npm i @types/jest` or `npm i @types/mocha`.ts(2582)
```

To be honest, this doesn’t really matter. I figured it would be nice to open the test files and not get screamed at with errors. JavaScript tooling still has a long way to go in some areas.

kleinfreund:
I put the Axe tests back in, but added the `.skip` string to the tests so that they’re not executed. They would fail in their current state.

I currently don’t see any way to get Axe Core to run in our test suite.

No matter what I try, `wrapper.vm.$el.children` is an empty collection so passing `wrapper.vm.$el` to `axe.run` produces this error:

```
No elements found for include in page Context
```

noopkat:
@kleinfreund it looks like you have to dismiss it to make it completely go away, but you can merge this without dismissing it all the same. I think this is ready, and I really appreciate your exploration into fixing axe-core. I can take on further research as I have an idea on how we could at least do some static analysis to get this started.

Thanks for all of the maintenance work you've been doing lately 💯 👑

knitcodemonkey:
Thank you, everyone, for looking into this. I’ll give it a second pass to see If there is any more I can add. The implementation was minimal, and I was hoping to spend time incorporating cypress testing with axe-core, but figured something was better than nothing at the time.

Issue #144, opened 2019-07-20
Optimise vendor bundle size 


noopkat:
Hello,

I did some analysis on what's causing our vendor bundle size to be so large.
The low hanging fruit I found was Chart.js. I pointed at the minified bundle instead of the full dist and saved over 200KB 😱 

The main difference is that the full dist includes moment.js and we don't use the dependent features in chart.js currently.

I think we can even go one better and look into using [Chartist](https://gionkunz.github.io/chartist-js/index.html) instead and knock that vendor bundle down even more.

If we're on board, I can work on this optimisation over the next couple of weeks.

kleinfreund:
Adding the following rule to the `plugins` key in the Webpack configuration shaves off the locales from Moment.js:

```
new webpack.IgnorePlugin(/^\.\/locale$/, /moment$/)
```

Adding the following entry to the configuration removes all of Moment.js but also breaks the `LineChartCard`:

```
externals: {
  moment: "moment"
}
```

Comparison of our bundle size with Chart.js:

| Technique    | Bundle size    |
|--------------|----------------|
| Before       | main (732 KiB) |
| IgnorePlugin | main (552 KiB) |
| externals    | main (500 KiB) |

Comparison of Chart.js’ and Chartist’s bundle size:

| Package  | Minified | Minified + Gzip |
|----------|----------|-----------------|
| Chartist | 38.6 kB  |  11.4 kB        |
| Chart.js | 385.1 kB | 112.1 kB        |

😱😱😱

@noopkat Did you try something else?

noopkat:
@kleinfreund I didn't ignore the bundle in the webpack config, rather I setup a new alias to resolve to the smaller chart.js bundle and that seemed to not break the linechart card's functionality. It was a shallow experiment to see how much moment.js was adding to the overall size.

Now that I have internet again, I just implemented the replacement of chart.js with chartist in a new branch and will open a pull request once I neaten it up a little bit.

PR #148, opened 2019-07-27
restore binding to cardform tile id field

noopkat:
A tiny regression that snuck in with some maintenance work 🔍 

I noticed it when I was working on #144. The cards weren't saving changes correctly and I traced it back to the id not binding correctly in the hidden tile id form field in the card settings component.

Let me know if this is a real bug this time! 😅

### DIFF
--- a/public/js/components/CardForm.vue
+++ b/public/js/components/CardForm.vue
@@ -3,7 +3,7 @@
     <h2>Settings</h2>
 
     <form @submit.prevent="onSubmit">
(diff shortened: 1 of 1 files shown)


kleinfreund:
Ah, ouch, indeed! I did a global string replacement of `v-bind:` to empty string (note how I included the `:` by accident). I noticed later and reverted the changes and this one I didn’t catch. Sorry! :+1:

PR #149, opened 2019-07-28
repair axe-core tests

### LINKED ISSUE #147: Axe tests are showing false positives
We use `axe-core` in the component unit tests to test for accessibility related issues with the code. However, they seem to always pass, even when the component contains accessibility violations.

Example: in `CardForm.vue`, un-nest the `<input>` from the `<label>` element [on line 10](https://github.com/noopkat/electric-io/blob/master/public/js/components/CardForm.vue#L10) so they are no longer associated, and run `npm test`. The accessibility test should fail, but it is passing.

We should fix the unit tests that use `axe-core` to ensure that they can fail on inaccessible components.

(Credit to @kleinfreund for spotting this - feel free to add anything I missed from your investigation into this issue! 😎)

noopkat:
Fixes #147 

I ended up installing [jest-axe](https://github.com/nickcolley/jest-axe) in order to resolve the axe-core test bugs in our suite. It did initially fail the card form component test because it didn't respect the nested label and field syntax we use. [I added in a `for` attribute](https://github.com/noopkat/electric-io/compare/fix-axe-tests?expand=1#diff-984fdc61423a9ada482bdb7fad7165f5R8) and sure enough the test started passing. So we definitely know it's now receiving the html output of the shallow rendered component instances 🤞 

Let me know what you think of this approach. I've been working for long hours today and am a bit fried so if this seems like the wrong approach I'm all ears 👂


public/js/components/specs/CardForm.spec.js
```diff
@@ -1,8 +1,9 @@
-import axe from "axe-core";
 import { shallowMount } from "@vue/test-utils";
+import { axe, toHaveNoViolations } from 'jest-axe';
 
 import CardForm from "../CardForm";
 
+
```
noopkat: sloppy sloppy I'll fix this whitespace

noopkat:
@kleinfreund thanks for fixing my linting - I'm having trouble getting husky to run on my pixelbook because of git version incompatibility problems 🙇‍♀

kleinfreund:
Splendid! If you’d see my browser search history, you would be confused as to how I could’ve possibly missed jest-axe, but here we are.

I’m absolutely fine with this. Seems to work well. :)

PR #150, opened 2019-07-28
migrate from chart.js to chartist

noopkat:
Somewhat addresses #144, by attacking the "low hanging fruit" 🍓 a.k.a. our charting library 📈 

| Vendor use | Vendor Bundle size | Minified + Gzip |
|------------|--------------------|-----------------|
| chart.js   | 629KB              | 198KB           |
| chartist   | 303KB              | 98KB            |

That is close to a 50% reduction in vendor bundle size 😱 🎉 

I was going to style the migrated chart to look exactly the same as the current chart, but I ended up liking the default look from chartist more. It looks more friendly because of the soft point interpolation, and the grid lines are helpful in my opinion for reading the chart. I still need to align the left axis to the left more by expanding the width of the chart container a little but that should be straightforward 🤞  Thoughts on the new aesthetic otherwise? 👀 

**Before:**

Screenshot 2019-07-28 at 21 56 56

**After:**

Screenshot 2019-07-28 at 11 07 30


noopkat:
@knitcodemonkey I know right? I was also surprised! Thanks for looking over this for me

PR #151, opened 2019-07-28
Clean up readme & friends

kleinfreund:
I cleaned up some documents in the project’s root directory:

- **License**:
  - This ain’t no Fog Creek project, right? Is this change fine?
- **Readme**:
  - Removed troubleshooting steps for Node and git installation as their respective installation tools have become pretty good. I don’t think their needed here anymore and covered better elsewhere.
- **Contributing**:
  - Mirror changes in README.md. We’re duplicating some content here, so I updated that file as well.


noopkat:
Beautiful! 

bitmoji

PR #153, opened 2019-08-04
Additional Frontend Component Code Coverage

jeffnhorner:
Hey Suz and noopcrew!

Slowly but surely trying to get code coverage across all of the frontend components. This pr gives additional coverage across 4 of the Vue components, one being the BaseCard which was a bit of work (and a lot of fun!). 

Screen Shot 2019-08-04 at 9 08 00 AM

The changes you see in the ButtonCard & ButtonSettings components themselves are actually related to the unit tests failing due to Axe finding a few accessibility violations - it was actually pretty sweet to see Axe working and doing it's job! 😄 

As always, there's definitely room for improvement and more work to do here, but we're making progress!

### DIFF
--- a/public/js/components/ButtonCard.vue
+++ b/public/js/components/ButtonCard.vue
@@ -1,10 +1,6 @@
 <template>
   <div>
-    <button
-      class="action-button"
-      @click="onClick"
-      v-bind:aria-label="ariaLabel"
-    >
+    <button class="action-button" @click="onClick">
       {{ tile.buttonText }}
     </button>
     <p :class="statusClass">
@@ -20,10 +16,6 @@ export default {
   props: {
     tile: {
       type: Object,
-      required: true
-    },
-    ariaLabel: {
-      type: String,
       required: true
     }
   },
--- a/public/js/components/ButtonCard.vue
+++ b/public/js/components/ButtonCard.vue
@@ -1,9 +1,12 @@
 <template>
   <div>
-    <button class="action-button" @click="onClick">
+    <button
+      class="action-button"
+      @click="onClick"
+      v-bind:aria-label="ariaLabel"
+    >
       {{ tile.buttonText }}
     </button>
-
     <p :class="statusClass">
       {{ statusText }}
     </p>
@@ -17,6 +20,10 @@ export default {
   props: {
     tile: {
       type: Object,
+      required: true
+    },
+    ariaLabel: {
+      type: String,
       required: true
     }
   },
--- a/public/js/components/ButtonSettings.vue
+++ b/public/js/components/ButtonSettings.vue
@@ -15,7 +15,7 @@
 
     <label>
       Call Type
-      <select id="callTypeSelect" name="callType" v-model="tile.callType">
+      <select id="callTypeSelect" v-model="tile.callType" name="callType">
         <option
           v-for="option in typeOptions"
           :key="option.value"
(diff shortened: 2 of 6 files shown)


public/js/components/ButtonCard.vue
```diff
@@ -1,9 +1,12 @@
 <template>
   <div>
-    <button class="action-button" @click="onClick">
+    <button
+      class="action-button"
+      @click="onClick"
+      v-bind:aria-label="ariaLabel"
```
kleinfreund: This is counterproductive. The button already has an accessible name in the form of its button text. Providing an `aria-label` attribute overrides this when announcing the button in screen readers. Also, no component in our code base passed `ariaLabel` to a `ButtonCard` so this effectively _removes_ the button’s accessible text.

Why did you add this? Was it due to Axe reporting a violation?

public/js/components/ButtonSettings.vue
```diff
@@ -15,7 +15,7 @@
 
     <label>
       Call Type
-      <select id="callTypeSelect" v-model="tile.callType" name="callType">
+      <select id="callTypeSelect" name="callType" v-model="tile.callType">
```
kleinfreund:
```suggestion
      <select id="callTypeSelect" v-model="tile.callType" name="callType">
```

I’d like to revert this change as there is a linter rule from the ESLint plugin for Vue marking this order of attributes as a warning.

If you use an editor like VS Code, you should be able to see the warnings. Since this was introduced very recently, you might run `npm install` on your branch to update the dependencies.

In any case, don’t worry about this, this is not important.

public/js/components/ButtonSettings.vue
```diff
@@ -27,20 +27,30 @@
       </select>
     </label>
 
-    <label v-if="tile.callType === 'method'">
+    <label for="deviceMethod" v-if="tile.callType === 'method'">
```
kleinfreund:
```suggestion
    <label v-if="tile.callType === 'method'" for="deviceMethod">
```

(For the same reason as in my previous comment about ESLint.)

jeffnhorner: Reverted 👍

public/js/components/ButtonSettings.vue
jeffnhorner: 10-4, reverted back to original attribute order.

public/js/components/ButtonCard.vue
jeffnhorner: Hey @kleinfreund, yep, that's exactly why I added it. Here's the violation that Axe reported: 

Screen Shot 2019-08-04 at 9 40 46 AM

Here's the Axe documentation it refers to help fix the violation: https://dequeuniversity.com/rules/axe/3.1/button-name?application=axeAPI

Using the title attribute may be the simplest solution here:

Screen Shot 2019-08-04 at 9 43 05 AM

Axe doesn't report any violations with this update - let me know what you think.

jeffnhorner: We could even add a prop for the title so we could be a bit more specific on the button title depending on how/where we're using the component - just a thought.

jeffnhorner:
> Hello there. :) Thank you for yet another improvements to our test suite. By now, we might actually call it that: A test suite.
> 
> I just have small remarks in a few places.

As always, thanks for your feedback and review @kleinfreund!

public/js/components/ButtonCard.vue
kleinfreund: Ah, then it’s very straightforward: In your tests for the ButtonCard, you want to provide a `buttonText` property on the `tile` object. Then, the button text won’t be empty. :)

jeffnhorner: Lol oh I gotcha. Thanks man! Pushing your requested changes now.

noopkat:
thank you for more tests 💯 

bitmoji

Issue #152, opened 2019-07-30
updated app screenshot for README


noopkat:
The screenshot of electric io in the README is looking noticeably out of date. I think it's time for a fresh one! 

If anyone wants to get creative, feel free to submit a new one. Having an example of all card types on there would be rad, similar to the existing screenshot. Perhaps keep the background non busy just to showcase the cool cards themselves.

PR #156, opened 2019-10-03
issues-152: Update screenshot for README

mpaarating:
Per #152, this PR updates the screenshot shown in the README.

I chose a green background because I think it is lovely.

[screenshot of electric io application](https://gyazo.com/c96868a43b7ae999c8a6aaafdcd532ad)


kleinfreund:
This looks goot to me; I just have a small nitpick: If you would be so kind, could you make sure that there is a tiny gap between the Cool Button and the Sticker Time cards? :)

Two bordered shapes barely overlapping

The buttons closely sticking together like this makes me a bit nervous.

mpaarating:
@kleinfreund good catch. I took the opportunity to somewhat better organize the windows compared to what I had before.

mtheoryx:
That looks awesome to me!

noopkat:
beautiful work, @mpaarating ! 
Thanks for the updated screenshot ✨ 

image

PR #158, opened 2019-10-05
Update all dependencies (minor & patch versions); npm audit fix

kleinfreund:
Packages, packages, packages.

- Updates all dependencies (happened to be no major version updates)
- Runs `npm audit fix` to fix vulnerability in a handlebars version (third party)


mpaarating:
Looks good to me! Thanks for always keeping up with this!

PR #159, opened 2019-10-15
Adds Alex

underscoredotspace:
Inspired by [PR 1337](https://github.com/facebook/react-native-website/pull/1337) for the React Native website.


CODE_OF_CONDUCT.md
```diff
@@ -1,3 +1,5 @@
+<!-- alex disable-->
```
underscoredotspace: Thought it was safe to ignore this file

CONTRIBUTING.md
```diff
@@ -162,7 +162,7 @@ Then, check out a new branch with `git checkout -b my-branch-name`
 
 Create your code / documentation changes in this branch, and commit when done. Once done, `git push fork your-branch-name`
 
-Once this is finished pushing, you can go to your fork on GitHub. It should now ask you right away if you want to create a pull request. Clicking that button should set you up with a text field just like when creating a new issue on GitHub. Fill it out and submit the pull request. Then, we can review it “Pull requests” in this original repository.
```
underscoredotspace: I know this isn't really a problem word in this case, but felt it was better to be able to leave the strong check than add an ignore to the file or sentence.

CODE_OF_CONDUCT.md
underscoredotspace: I should clarify: the content causes alerts (4 I think) which are obviously not needed. We should be able to control this file ourselves.

underscoredotspace: I think this is going to be the only file that would ever be ignored, which is why I chose this option.

It means a lot to have had your eyes on this @ricknahlonii. I'm greatful that you took the time to do this in the React Native repo. It's something I use almost every day at work - thank you 🙌

noopkat:
Hey @underscoredotspace sorry I had to disappear from this repo for a while and focus on avrgirl. This change is awesome and I didn't know about Alex until now and I really like it after playing with it briefly. Merged! ✅ 

Thank you to @kleinfreund and @rickhanlonii for reviewing this 💝 🙇‍♀ 

bitmoji

Issue #165, opened 2020-02-02
Managing Dependencies with Renovate


lannonbr:
So with #161 just merged, [Renovate](https://github.com/renovatebot/renovate) was added to this project. It will work alongside your package.json and update dependencies automatically and send PRs. This on top of Azure Pipelines will allow a nice system so when Renovate creates a new PR, the CI will run and then we can have some clarity that the dependency update is sucessful.

That said, with about 50 direct deps / devdeps, this can possibly push a lot of PRs. Renovate has a lot of [Config options](https://docs.renovatebot.com/configuration-options/) so I was going to open this issue just to get a discussion rolling on how renovate should function on this project.

MichaelE1:
Great issue. I had a quick look in Azure Pipelines and it seems that Suz has already been API rate limited 😬. https://dev.azure.com/noopkat/electric-io/_build/results?buildId=212&view=results

I don't know much about Renovate, but it seems possible to [group all updates together](https://docs.renovatebot.com/presets-group/#groupall). That could help reduce the API calls and PRs (& notifications!).

benhalverson:
I like the idea of grouping the updates in renovate that's the one thing that annoys me in my repos

PR #188, opened 2020-02-11
Group renovate dependency updates

mikeesto:
Attempts to resolve https://github.com/noopkat/electric-io/issues/165 by grouping all dependency updates together.

While perusing the renovate docs I noticed that it's also possible to [update everything together once a month](https://docs.renovatebot.com/presets-config/#configsemverallmonthly). I haven't included that in this PR but I think it's worth considering.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,5 +1,5 @@
 {
-  "extends": ["config:base"],
+  "extends": ["config:base", "group:all"],
   "automerge": true,
   "major": {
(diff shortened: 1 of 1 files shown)


noopkat:
@mikeesto thanks for seeing this feature need and for @lannonbr opening the issue, I was a little snowed under there with all of the individual PRs it was opening 😅 

bitmoji

Issue #196, opened 2020-03-05
Discuss Azure Event Hub SDK 5.0


tkoster:
Hi again Suz and the noopcrew! :wave:

I thought I would create this issue to discuss the breaking changes caused by the upgrade to `@azure/event-hubs@5` proposed by renovate in #192 , as I did some investigation into this that I'd like to share.

Feel free to elaborate and make corrections :sweat_smile:

## Background

The :zap: electric-io server uses two libraries to access the IoT Hub:

1. the [Event Hub SDK](https://www.npmjs.com/package/@azure/event-hubs) is used to subscribe to device-to-cloud messages to show on the dashboard
2. the [IoT Hub SDK](https://www.npmjs.com/package/azure-iothub) is used to list devices and send cloud-to-device messages

This issue is about the first one. At the time of writing, electric-io uses `@azure/event-hubs@2.1.1`. The latest version is `@azure/event-hubs@5.0.1`.

Of the many changes, two seem to affect electric-io adversely: the splitting of `EventHubClient` into `EventHubConsumerClient` and `EventHubProducerClient`, and the discontinuation of support for IoT Hub connection strings when creating Event Hub clients.

## Splitting client classes

The splitting of the client classes should hopefully require only fairly mechanical changes in the electric-io server. Most importantly, this should present no breaking change to electric-io users.

## Connection strings

Both IoT Hub and Event Hub use connection strings, but they are incompatible.

However, older versions of the Event Hub SDK `@azure/event-hubs` include a very convenient function, `EventHubClient.createFromIotHubConnectionString` that allowed you to use an IoT Hub connection string to connect to the Event Hub compatible endpoint in the IoT Hub transparently. Electric-io takes advantage of this to simplify the user experience: users only need one connection string - the IoT Hub connection string.

See [lib/liveHub.js](https://github.com/noopkat/electric-io/blob/554636cb806ced0f7ec31256fef86485e0c2eec1/lib/liveHub.js#L25).

Unfortunately, this function was removed from the SDK in late 2019, in https://github.com/Azure/azure-sdk-for-js/pull/5311.

[...]

PR #197, opened 2020-03-08
Update event-hubs to 5.0.1

tkoster:
Hi @noopkat and the electric-io community! :wave: :coffee: :v:

This pull request upgrades `@azure/event-hubs` to the latest version on npm: 5.0.1.

As mentioned on #196, we can no longer rely on the fantastically convenient static constructor `EventHubClient.createFromIotHubConnectionString`. This PR only goes as far as making electric-io functional again with version 5.0.1 of `@azure/event-hubs`. It does not address the connection strings issue and therefore electric-io now requires two connection strings in this PR: :cry: 
- the IoT Hub connection string which you already have, and
- the connection string of its [Event Hub-compatible endpoint](https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-messages-read-builtin).

As Suz already lamented, this is a **breaking change** for electric-io users :sob: 

As such, I've included some things in the `server.js` and updated the `README.md` to ease the pain until we decide what to do about #196.

On the other hand, if the breaking changes prevent a merge then this PR can serve as a starting point where electric-io is at least functional with this version of `@azure/event-hubs`.

## Breaking changes

1. The `CONNECTION_STRING` environment variable (holding the IoT Hub connection string) is **renamed** to `IOTHUB_CONNECTION_STRING`.
2. **Added** a new environment variable `EVENTHUB_CONNECTION_STRING` for the connection string of the [Event Hub-compatible endpoint](https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-messages-read-builtin) of the IoT Hub.

## Migration

If the server detects that `CONNECTION_STRING` is present but not `IOTHUB_CONNECTION_STRING`, it prints a message telling the user that this has changed and prints a link to the README.

To fix your configuration:

1.  Edit the `.env` file (or your Docker compose file) and rename the `CONNECTION_STRING` variable to `IOTHUB_CONNECTION_STRING`.
2.  Go to the Azure Portal, navigate to your IoT Hub, open the "Built-in endpoints" blade and copy your "Event Hub-compatible endpoint".
    azure-portal-iothub-eh-endpoint
3.  Back in `.env`, add a new variable `EVENTHUB_CONNECTION_STRING` and paste the Event Hub-compatible endpoint connection string copied in step 2.

## Message contract coupling

During this upgrade, I noticed the electric-io server sends the message object received from the Event Hub directly to the websocket as-is. This couples the UI component that is subscribed to the websocket to the Event Hub client library data structures.

For example, as part of the changes in `@azure/event-hubs`, the `annotations` property on this was renamed. This broke the event handler in `App.vue`.

In future, we may like to consider decoupling the message contract between the server and the browser from the Event Hubs types so that changes in the Event Hub client do not affect the browser app. This would probably be required to encourage contributions for alternative hubs too.

---

It also looks like prettier had some fun in the markdown files.

### DIFF
--- a/.env
+++ b/.env
@@ -5,6 +5,7 @@
 
 # note: .env is a shell file so there can't be spaces around '=
 
-CONNECTION_STRING=''
+IOTHUB_CONNECTION_STRING=''
+EVENTHUB_CONNECTION_STRING=''
 CONSUMER_GROUP=''
 SIMULATING=true
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -5,13 +5,13 @@ Hey! Thanks for your interest in contributing to this project! I really apprecia
 This page should get you up and running with how to run locally and how to adhere to this project’s style guide also.
 
 1. [Installation](#installation)
-    - [With a real Azure IoT Hub](#with-a-real-azure-iot-hub)
-    - [With the device and hub simulator](#with-the-built-in-device-and-hub-simulator)
+   - [With a real Azure IoT Hub](#with-a-real-azure-iot-hub)
+   - [With the device and hub simulator](#with-the-built-in-device-and-hub-simulator)
 2. [Style guide](#style-guide)
-    - [Tests](#tests)
-    - [Mobile support](#mobile-support)
+   - [Tests](#tests)
+   - [Mobile support](#mobile-support)
 3. [Pull requests](#pull-requests)
-    - [Avoiding merge conflicts](#merge-conflicts)
+   - [Avoiding merge conflicts](#merge-conflicts)
 
 ## Installation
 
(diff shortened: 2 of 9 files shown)


tkoster:
I'll try and say hi tonight but no promises as the stream starts at 4am in my timezone :zzz: . I'm `codeandcoffee` in Twitch chat (apparently I'm not the only contributor who likes these two things :smile: )

tkoster:
Hi @noopkat :sparkles:, I just wanted to let you know that I don't want you to feel any pressure or urgency to deal with this pull request (unless a vulnerability forces our hand of course).

I support your decision if you want to hold on to this PR a while longer while we take some more time to think about how we can avoid inconveniencing users with this breaking change, which I agree is right up there as far as inconvenient breaking changes goes. This change doesn't even have any cool new features that we can use to bribe users :sweat_smile:

Keep up the good work :v::coffee:

tkoster:
Be sure to check out #198 for an alternative option.

tkoster:
I'm closing this pull request - no need for it any more since #198 was merged :tada:

PR #198, opened 2020-03-13
Update event-hubs to 5.0, compatible edition

tkoster:
Hi Suz! :v: :coffee:

This pull request, like #197, upgrades @azure/event-hubs to version 5.0.1, and is an **alternative** to #197.

As discussed in issue #196, the `EventHubClient.createFromIotHubConnectionString` function is removed from the Event Hubs client library. :sob:

To avoid inconveniencing users with the burden of a second connection string (the connection string of the Event Hub-compatible endpoint), this pull request includes a reimplementation of that function's behaviour.

Hopefully we haven't overlooked a critical reason why doing this might be a bad idea and why it was removed from the Azure library.

## Breaking changes

Hopefully none! :tada:

## Approach

I was encouraged by the short Python program in the "Receive telemetry messages (service client)" section in [the IoT Hub AMQP documentation](https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-amqp-support#receive-telemetry-messages-service-client) that performs the same trick :snake:

To recap, the trick is this:

1. Connect to the IoT Hub with the AMQP protocol
2. Authenticate
3. Send a message to the management interface, which causes a link redirect error to the Event Hub compatible endpoint where the real management interface is
4. Parse the link redirect error and construct the full Event Hub connection string

This is implemented in `lib/azureIotHubManagement.js` and used in `liveHub.js` at startup.

Authentication is performed using the `CbsClient` class from the [@azure/core-amqp](https://www.npmjs.com/package/@azure/core-amqp) library, which I believe is also used by the IoT Hub, Event Hub and Service Bus libraries.

The function then sends a message to cause the AMQP link redirect error that reveals the missing parts we need to construct the Event Hub connection string.

## Testing different connection strings

I use Shared Access Policy type connection strings for IoT Hub which I got by following the getting started instructions in the Electric-IO README. I use a custom policy - not the default "iothubowner" policy. This is the only type of connection string I have tested this with. Despite appearances, I am actually not very familiar with IoT Hub or Event Hub and don't know whether or not other exotic configurations exist where this function does not work and further testing by the electric-io community might be useful.

---

I hope you don't mind some `async`/`await` :sweat_smile:
TypeScript would have been nice too for exploring the unfamiliar libraries :stuck_out_tongue:

### DIFF
--- a/.env
+++ b/.env
@@ -5,7 +5,6 @@
 
 # note: .env is a shell file so there can't be spaces around '=
 
-IOTHUB_CONNECTION_STRING=''
-EVENTHUB_CONNECTION_STRING=''
+CONNECTION_STRING=''
 CONSUMER_GROUP=''
 SIMULATING=true
--- a/.env
+++ b/.env
@@ -5,6 +5,7 @@
 
 # note: .env is a shell file so there can't be spaces around '=
 
-CONNECTION_STRING=''
+IOTHUB_CONNECTION_STRING=''
+EVENTHUB_CONNECTION_STRING=''
 CONSUMER_GROUP=''
 SIMULATING=true
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -5,13 +5,13 @@ Hey! Thanks for your interest in contributing to this project! I really apprecia
 This page should get you up and running with how to run locally and how to adhere to this project’s style guide also.
 
 1. [Installation](#installation)
-   - [With a real Azure IoT Hub](#with-a-real-azure-iot-hub)
-   - [With the device and hub simulator](#with-the-built-in-device-and-hub-simulator)
+    - [With a real Azure IoT Hub](#with-a-real-azure-iot-hub)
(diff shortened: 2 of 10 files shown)


noopkat:
Hi @tkoster this is absolutely bloody amazing. I'm going to take a closer look at this, and might connect with an ex-colleague to ensure this can be a pretty reliable workaround longterm. I'm leaning towards merging this very soon though, and then prioritising working on MQTT support which will also help with ensuring different ways to connect to IoT Hub as a backup.

🙇 🎉

tkoster:
Hi @noopkat, that is wise, considering we're definitely in "workaround" territory with this. If I didn't see it published in official documentation (in the Python program), I wouldn't have attempted this.

Of course, your contact may instead confirm that the 2.x series of @azure/event-hubs is still supported and we should just upgrade to 2.1.4 instead and tell the renovate bot to chill!

noopkat:
@C0ffeeCode THIS RULES. Finally got to verify it on my end and thanks for the live help!

a very smiley lab retriever dog

PR #202, opened 2020-04-11
Fix card settings not always being saved correctly

kleinfreund:
Good day! :)

Fixing a bug you ran into last week on stream. I swear I fixed that one before, but that was obviously not the case.

I’m also the one who introduced the bug in the first place. It happened during the re-implementation of the card dragging logic.

## Changes

- Fixes a bug where the current content of a card settings form was lost on certain actions (e.g. clicking outside the card or clicking the save button). This made it look like the data wasn’t being saved. What actually happened was the form data being reset just when save was clicked.
- Adds an early return to the callback processing socket messages. It currently throws on a property systemProperties being accessed when it’s not always defined.

### DIFF
--- a/public/js/components/App.vue
+++ b/public/js/components/App.vue
@@ -169,6 +169,10 @@ export default {
 
       const socket = io();
       socket.on("message", message => {
+        if (!("systemProperties" in message)) {
+          return;
+        }
+
         const deviceId =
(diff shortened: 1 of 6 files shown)


noopkat:
Thanks @kleinfreund for fixing this bug so quickly 😻 

terrific

PR #203, opened 2020-04-12
Add ElectricToaster component

### LINKED ISSUE #126: new UI for dashboard saves
What I would like to see is an error [toast](https://www.patternfly.org/pattern-library/communication/toast-notifications/) that:

+ [ ] shows an error if dashboard doesn't save correctly
+ [ ] has a long timeout for a11y reasons (or preferably have a way to dismiss it via user gesture)
+ [ ] is a live region / role of alert so that screenreader users can hear the error when it happens

kleinfreund:
Closes #126.

Adds a new component called ElectricToaster ⚡🥪. It’s a smol toast notification component with the following features:

- Toasts are provided as an array of objects. The parent component (in electric-io, that’s App.vue) is responsible for providing the toast objects.
- Toasts can be created with the properties “content”, “shouldAutoDismiss”, and “autoDismissTimeoutInSeconds”.
  - `content`: Required. This property must contain the content of the toast as a string.
  - `shouldAutoDismiss`: Optional. Controls whether this toast will be automatically dismissed. When not provided, the ElectricToaster’s default will be used (currently set to an annoyingly long 10 seconds)
  - `autoDismissTimeoutInSeconds`: Optional. Controls the timeout after which a toast will be dismissed. If `shouldAutoDismiss` is `false`, this property has no effect.
- Active (i.e. not dismissed) toast messages are automatically placed in a visually-hidden ARIA live region with the attributes `role="region"` and `aria-live="polite"`. This is intended to cause screen readers to announce the content of the live region as soon as any in-progress announcements are finished.

**Screenshot**

Screenshot from 2020-04-12 11-22-18

<details>
<summary><b>Demo GIF</b></summary>

electric-toast

</details>

## To do

- [x] Use ElectricToaster for error messages from the API endpoints. This will need some re-writing of the the `postApiDataTo` function in `public/js/lib/configuration.js` and possibly other places. Maybe that’s a good task to handle on-stream? At least we can discuss what we want to have there. 🤓⚡💾

### DIFF
--- a/lib/routes.js
+++ b/lib/routes.js
@@ -18,31 +18,75 @@ function injectRoutes({ io, hubService }) {
 
   routes.get("/api/dashboard", function(req, res) {
     debug("dashboard requested");
+
+    const localDashboardPathSegment = path.join(".data", "dashboard.json");
+
     fs.readFile(
-      path.join(__dirname, "..", ".data", "dashboard.json"),
+      path.join(__dirname, "..", localDashboardPathSegment),
       { encoding: "utf8" },
       (error, data) => {
         debug(data, error);
-        if (error) return res.sendStatus(500);
-        let jsonData = JSON.parse(data);
-        jsonData.dashboard.editMode = editMode;
-        res.send(JSON.stringify(jsonData));
+
+        if (error) {
+          const message = `Could not read from file “${localDashboardPathSegment}”.`;
+
+          return res.status(500).send({
+            data: { message }
+          });
+        } else {
+          try {
+            let jsonData = JSON.parse(data);
+            jsonData.dashboard.editMode = editMode;
+
+            return res.status(200).send(JSON.stringify(jsonData));
+          } catch (error) {
+            const message = `Could not parse the contents of “${localDashboardPathSegment}” as JSON.`;
+
+            return res.status(500).send({
+              data: { message }
+            });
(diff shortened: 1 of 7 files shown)


kleinfreund:
## Error handling

In order to allow the frontend to consume backend errors in straightforward mannor, I rewrote most of the code related to sending API responses via express and the API wrappers in the frontend.

Requests resulting in a non-success status code will now always return a response body of the following structure:

```
{
  data: {
    message: string
  }
}
```

If this structure is present, the API wrappers in the frontend will throw `data.message` as the content of a new error. Otherwise, the response’s status text (e.g. “Internal Server Error”) is used.

In places where that’s possible, the frontend will now notice these error messages and show a toast notification. For example, if for some reason, the call to `fs.writeFile` fails upon trying to save the dashboard settings, a notification reading “Could not write to dashboard.json” will be created. Likewise, if the dashboard.json file is not valid JSON, the dashboard won’t load, but now, there will be a toast notification letting the user know (and it won’t automatically be dismissed).

Screenshot from 2020-04-12 13-47-52

Also, async/await is beautiful!

noopkat:
I am so glad we finally have some toasts and I LOVE the name ⚡ 🥪 
Thank you @kleinfreund !!

wiggle penguin

PR #206, opened 2020-04-12
refactor: split off services from route definitions and add tests

### LINKED ISSUE #76: Add tests to the express server! Yay!
There are a few TODOs in the index.js of the express server file that are very valid!

```javascript
// TODO: break this file down into pieces so it's easier to digest
// ideas:
// 1. load routes from routes file
// 2. put routes handers in a file
// 3. put iot related work in a file
//
```

However, you cannot refactor untested code. That's just moving stuff arround. So this issue is to put some tests around the routes at least. THEN we can start moving stuff around :)

kleinfreund:
This is a general refactoring of the backend route definitions and their dashboard file and IoT Hub related service methods.

- A new HubService file was created in `lib/services/hub.service.js`. It exports an object that holds the following methods:
  - `getDeviceIds`: Retrieves a list of device IDs from the IoT Hub Service.
  - `callDeviceMethod`: Calls a method with a certain method name and payload on a specific device (as identified by a device ID).
  - `callDeviceMessage`: Sends a message to be called on a specific device (as identified by a device ID).
- A new DashboardService file was created in `lib/services/dashboard.service.js`. It exports an object that holds the following methods:
  - `getDashboardSettings`: Reads the dashboard settings from `.data/dashboard.json`.
  - `saveDashboardSettings`: Saves a dashboard settings object to `.data/dashboard.json`.
- Restructures the server’s route definitions in `lib/routes.js`. All route definitions now follow a common pattern of calling and awaiting the value of a service method (see the two points above). If the value is awaited successfully, a response with status code 200 is sent. The response’s payload will the resulting value of the service method. If an error is thrown while awaiting the the result of the service method, a response with status code 500 is sent instead.

  **Note**: There is currently an exception for `HubService.getDeviceIds` as this method currently never rejects with an error.
- Adds a new test file `lib/routes.test.js`. It has complete test coverage for all route definitions.
- Adds a new test file `lib/services/dashboard.service.test.js`. It has complete test coverage for the `DashboardService` methods.
- Adds a new test file `lib/services/hub.service.test.js`. It has complete test coverage for the `HubService` methods.
- Adds a new test file `lib/utilities/createDashboardFileIfItDoesNotExist.test.js`. It has complete test coverage for the utility function `createDashboardFileIfItDoesNotExist`.
- Renames the file `.data/dashboard.json.sim` to `.data/dashboard.example.json` (primarily to allow programs to make use of the .json file extension).
- Fixes the some endpoint’s responses not having the correct content type (was returning `text/html`; is now returning `application/json; charset=utf-8` when returning JSON).

## To do

- [ ] Someone with a working IoT hub setup should test if the client-server interactions relating to devices still work.

Closes #76.


noopkat:
this is a really lovely and detailed comb through the backend code base in order to make things much much better. Thank you for having the patience to sit down and do this 🙇 

turtle enjoying a toothbrush scrub on its shell

tkoster:
Hi @kleinfreund,

I encountered the 404 issue that @noopkat had on the weekend. I took the opportunity to investigate it. The stack is very deep into promises and confusing, but I believe this is what is happening:

Although the HTTP server is not started, `server.js` does attempt to connect to the real IoT Hub if your `.env` file says so. In my case, my connection string was wrong and it was unable to initialise the live hub. I can reliably reproduce and fix the issue by switching between my `.env` file and the default `.env` file.

---

In the `startHubServer` function, it initialises the hub before registering any routes with express (presumably for the good reason that the handlers need to be injected with the appropriate hub service). If initialising the hub throws an exception, the routes are not initialised.

A quick fix might be to completely ignore `.env` and force simulation mode when `NODE_ENV` is `test`, similar to how the HTTP server start is elided.

However, the root cause seems to be that `server.js` is not really a reusable module that one can "require", but rather is the "main" of the Node app. What happens when you "require" it is quite nondeterministic: it changes its behaviour drastically depending on the environment.

A more permanent fix might be to reduce the nondeterminism in these functions by changing references to the environment variables into formal parameters that can be controlled by the tests, and then move the functions into a module that can be required by the tests without these side-effects.

kleinfreund:
@tkoster Ah, I see. That makes sense now, finally. And I couldn’t have possibly ran into it either, unfortunately. Would you mind creating an issue with exactly the content of your message here? I think it’s a good idea to enforce the tests running under simulating conditions for everything that doesn’t require a non-simulated hub for testing.

tkoster:
@kleinfreund will do.

noopkat:
thanks for looking into this @tkoster this all makes a lot of sense 👍

Issue #207, opened 2020-04-14
Value-dependent colour gradient


tkoster:
How do folks feel about a colour gradient mode for the **number card**?

![colourgradient.gif](https://user-images.githubusercontent.com/9063335/79248576-58175980-7ebf-11ea-90be-9972c4eb8c19.gif)

I'm a bit stuck for ideas on how to fit it into the settings view though. Both horizontal and vertical layouts make the card much too big imo.

![colourgradientsettings.png](https://user-images.githubusercontent.com/9063335/79248921-d673fb80-7ebf-11ea-97c7-8f410047c5c2.png)

kleinfreund:
We can think about hiding the color picker by default or at least one of them. The no-fuzz approach would be wrapping it in a details/summary element combo.

Also, Suz’s room is almost boiling if those are degree celsius. :hot_face: :hot_pepper:

tkoster:
It's not over by much :joy:
![image](https://user-images.githubusercontent.com/9063335/79254714-7b92d200-7ec8-11ea-9ee7-44c4f47b34d3.png)

tkoster:
Do you mean something like this?

![electricio-gradient-settings-2](https://user-images.githubusercontent.com/9063335/79446244-43eb6d80-8021-11ea-8040-4c30762baebb.gif)

- one colour picker is visible
- visually, the picker displays/sets whichever colour (low or high) that last had focus
- when tabbing, the fields are visited in this order: low value, low colour, high value, high colour

kleinfreund:
Do I understand the GIF correctly that the color picker is made visible based on which of the two value text input fields was focused last?

I think that’s not super clear to someone who doesn’t know. There should be a clear interaction that toggles between the two color pickers. If I’m satisfied with the current setting of the “high value” field, I might never click into it and realize that I can change its associated color.

Let me try to illustrate what I suggested above with the details/summary elements:

---

## Settings

[the first few form controls]

Low value
[40_________]

<details>
<summary>Select low value color</summary>
[Low value color picker here]
</details>

High value
[80_________]

<details>
<summary>Select high value color</summary>
[High value color picker here]
</details>

[Save]
[Cancel]

---

Both color pickers would then be hidden from view by default.

tkoster:
You are correct. I'm trying to balance screen real estate with the desire to have low number of clicks/keyboard interactions to get to specific controls, and I agree that clarity if often the thing that is traded away when optimising for those.

I'll try your suggestion and some more options.

[...]

PR #209, opened 2020-04-23
Colour gradient for number card

tkoster:
Hi @noopkat and Electric I/O fans :wave: :coffee:

This change enhances the **number card** with a new gradient colour mode, implementing the fly-out design in #207.

The colour of the number is a blend of two colours where the ratio is determined by the value of the number.

In the example below, the gradient is set to blue at 40 and pink at 60. If the value is less than or equal to 40, the number is shaded blue. If the value is greater than or equal to 60, the number is shaded pink. Between 40 and 60, the number is shaded in a blend of blue and pink in a ratio that depends on how close to each number the value is.

screenshot of number card with colour gradient

## Card settings

The layout of the original single colour mode is unchanged.

The colour pickers for the new gradient mode appear in a fly-out card to save space. They appear when clicking the button for the colour you wish to change. The buttons are shaded in the colour of that setting.

The fly-out is not modal - the rest of the form can still be interacted with regardless of whether the fly-out is visible or not. Dismissing the fly-out is not necessary to save or cancel the form.

electricio-number-gradient-color-settings

The settings for the single colour mode are independent of the settings for the colour gradient mode.

## Approach

A lot of work was reused from the colour conversion functions and existing colour picker. Thanks contributors! 🥇 

The colour is calculated by interpolating linearly in hue-saturation-value (HSV) coordinates between the two colours, where hue is an angle interpolated either clockwise of counterclockwise depending on which distance is closer. The blend ratio is determined by the current value of the card.

I don't believe there is a "best" colour space for interpolation, but HSV seems to give adequate results. Luckily, because the colour picker itself shows a hue scale, you can easily predict which colours will appear in the blend by looking at the sequence of colours on this scale. In the screenshot below the colour picker's hue scale shows that interpolating between green and blue would include cyan colours:

interpolating between green and blue would include cyan colours

Alpha is interpolated as well. I'm not sure yet if this is useful :grin:

alpha is interpolated as well

## Dashboard schema migrations

As this change adds new properties to the dashboard that require sensible defaults, I added a very basic dashboard migration feature. When the server starts, it checks a `version` property and upgrades the dashboard if the value of the property is lower than the current version. (A dashboard with no `version` property is at version "0".)

Dashboard settings migrations are documented in more detail in `CONTRIBUTING.md`.

## Other changes

Cards that are flipped over to their settings now have a higher `z-index` than cards that are face up. This addresses the issue where settings can be hidden behind other cards.

My last PR about Event Hubs broke the simulation mode, which I've briefly fixed here too.

### DIFF
--- a/.data/dashboard.blank.json
+++ b/.data/dashboard.blank.json
@@ -5,7 +5,7 @@
     "tiles": [],
     "title": "⚡electric io",
     "bgImageUrl": "",
-    "bgImageRepeat": "true",
-    "version": 1
-  }
+    "bgImageRepeat": "true"
+  },
+  "version": 1
 }
--- a/.data/dashboard.blank.json
+++ b/.data/dashboard.blank.json
@@ -3,7 +3,7 @@
     "blockSize": [250, 200],
     "bgColor": "papayawhip",
     "tiles": [],
-    "title": "\u26a1electric io",
+    "title": "⚡electric io",
     "bgImageUrl": "",
     "bgImageRepeat": "true",
     "version": 1
--- a/.data/dashboard.blank.json
+++ b/.data/dashboard.blank.json
@@ -5,6 +5,7 @@
     "tiles": [],
     "title": "\u26a1electric io",
     "bgImageUrl": "",
-    "bgImageRepeat": "true"
+    "bgImageRepeat": "true",
+    "version": 1
   }
 }
--- a/.data/dashboard.example.json
+++ b/.data/dashboard.example.json
@@ -76,7 +76,7 @@
         "id": "88e12934-c628-48cb-8b84-b16c2e311105"
       }
     ],
-    "title": "⚡electric io",
-    "version": 1
-  }
+    "title": "⚡electric io"
+  },
+  "version": 1
 }
--- a/.data/dashboard.example.json
+++ b/.data/dashboard.example.json
@@ -24,7 +24,7 @@
       },
       {
         "title": "Study Temperature",
-        "type": "lineChart",
+        "type": "line-chart",
         "lineColor": "#FF6384",
         "deviceId": "AZ3166",
         "property": "temperature",
@@ -34,7 +34,7 @@
       },
       {
         "title": "Jenn's Coolness Factor 😎",
-        "type": "lineChart",
+        "type": "line-chart",
         "lineColor": "#6b66ff",
         "deviceId": "Jenn",
         "property": "coolness",
@@ -59,7 +59,7 @@
       },
       {
         "title": "Noise 🙉",
-        "type": "lineChart",
+        "type": "line-chart",
         "lineColor": "#60dd4d",
         "deviceId": "Tessel2",
         "property": "sound",
--- a/.data/dashboard.example.json
+++ b/.data/dashboard.example.json
@@ -50,7 +50,12 @@
         "property": "humidity",
         "position": [296, 465],
         "size": [1, 1],
-        "id": "101ca6cb-ba75-4e5c-9db6-d29ba7ab9f51"
+        "id": "101ca6cb-ba75-4e5c-9db6-d29ba7ab9f51",
+        "textColorMode": "single",
+        "lowValue": 20,
+        "lowTextColor": "#1282baff",
+        "highValue": 40,
+        "highTextColor": "#cb1475ff"
       },
       {
         "title": "Noise 🙉",
@@ -70,6 +75,7 @@
         "size": [1, 1],
         "id": "88e12934-c628-48cb-8b84-b16c2e311105"
       }
-    ]
+    ],
+    "version": 1
   }
 }
--- a/.data/dashboard.example.json
+++ b/.data/dashboard.example.json
@@ -76,6 +76,7 @@
         "id": "88e12934-c628-48cb-8b84-b16c2e311105"
       }
     ],
+    "title": "⚡electric io",
     "version": 1
   }
 }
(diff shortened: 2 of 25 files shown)


kleinfreund:
Hello there. 👋

This looks very cool. I’ve checked out your MR locally and looking at :sparkle: things :sparkles: now.

I’m just going to leave a couple of my thoughts here:

## Dashboard migrations

- I like the concept of versioned dashboard files. I think it makes a bit more sense if the version field is put in the outer-most JSON object of the file rather than in the `dashboard` object. What do you think about that? Ideally, anything in that file could be migrated to a new version if the underlying schema is changed. I do realize that this would require a bit of rewiring in the migration logic to not only be conerned with the `dashboard` field. Is there a particular reason to limit the migration functionality to that field?
- The newly added functions in `public/js/lib/configuration.js` should be placed in its own file (e.g. `public/js/lib/dashboard-migrations.js`) in my opinion. And to make the separation here a bit more obvious, I’d rename the `configuration.js` file to something like `api-adapter.js` or `api-wrapper.js`.
- In the `dashboardMigrations` array (a good idea to store the like this), I think I would only reference function names instead of defining them there, too. So for the current and only migration function, I’d define a function called `migrateDashboardVersion0To1` and reference it inside `dashboardMigrations`. One advantage of this is that it makes spotting incorrect migration chains easier. If I add a function `migrateDashboardVersion1To2` _before_ `migrateDashboardVersion0To1`, it’s easier to see how that is incorrect.

## Miscellaneous

- Thank you for adding tests. :+1: 
- Makes me very happy that the color conversion functions are being re-used like that.
- I pray that at the pearly gates, I’m forgiven for introducing this atrocity:

  ```javascript
  document.head.style.color = this.color;
  const rgbValue = getComputedStyle(document.head).color;
  ```

---

That’s it for now. Looking forward to seeing this on stream. 🙌

tkoster:
@kleinfreund:
> the version field is put in the outer-most JSON object of the file rather than in the `dashboard` object. ... Is there a particular reason to limit the migration functionality to that field?

I didn't notice this outer object was here. Now I see the `GET` and `POST` of the API route and the `getDashboardSettings` and `saveDashboardSettings` functions are asymmetrical: the gets return the outer object but the saves want the inner object. :thinking:

Because the code for loading and saving the settings seemed to be made just for dashboard settings, the thought that other things might end up in here did not cross my mind. See the function names (e.g. `saveDashboardSettings`), the API routes (`/api/dashboard`) and the filenames (`dashboard.json`). So.... yagni?

If you want to know more about my reasoning for why the migration is done in the Vue app and not in the server: Doing it this way might sound odd without context, but looking at the general structure of the app, all the concepts about cards and the dashboard settings are defined in the Vue app. The dashboard settings are completely opaque to the server. In my mind the server's responsibility is to deal with the IoT Hub and provide persistence for the dashboard settings, so locating the dashboard upgrade code near where the dashboard concepts are defined made more sense. There are still technical reasons for doing the migration in the server though, such as when multiple users concurrently edit the same dashboard, but I didn't think Electric I/O for set up for this use case.

tkoster:
> The newly added functions in public/js/lib/configuration.js should be placed in its own file

> In the dashboardMigrations array (a good idea to store the like this), I think I would only reference function names instead of defining them there

Good points, I'll do this.

.data/dashboard.example.json
```diff
@@ -70,6 +75,7 @@
         "size": [1, 1],
         "id": "88e12934-c628-48cb-8b84-b16c2e311105"
       }
-    ]
+    ],
```
kleinfreund, line 78: I noticed that the example dashboard is missing a title value.

```suggestion
    ],
    "title": "⚡electric io",
```

.data/dashboard.blank.json
```diff
@@ -5,6 +5,7 @@
     "tiles": [],
     "title": "\u26a1electric io",
```
kleinfreund, line 6:
```suggestion
    "title": "⚡electric io",
```

tkoster:
> In the future, the endpoint for saving the dashboard should also expect the wrapper object
> ... But I don’t think this should block these changes from being merged

In that case we should make this symmetrical and move the version property up to the outer object before merging this. Doing this later would mean the migrations need to be able to check for the version in either the inner or outer object, plus an added migration to move the version number outwards, which I would rather not do. I'll see if I can get this done before tonight.

tkoster:
Actually, while the server side changes are small, adding a wrapping object requires major surgery in `App.vue`. To avoid unsightly `this.dashboard.dashboard` everywhere, I should rename it to something like `this.settings.dashboard`, but then several functions like `saveDashboard` should be renamed too. Also, the immutable updates (`Object.assign({}, this.dashboard, ...)`) would require an additional level of depth.

I feel all this is turning into too much for this PR.

If your concern is basically that the `version` property is not at the top of the file, then the other way to make everything symmetrical is to _remove_ the outer wrapper when returning the dashboard JSON to the Vue app. In other words, make the current dashboard object (with `tiles` and `blockSize` properties) the outermost object seen by the Vue app. That means I need to move the migration into the server, but oh well. To me this might be the path of least resistance for making the load/save symmetrical with a `version` property on the top level object in the dashboard file.

What do you think @kleinfreund ?

kleinfreund:
@tkoster I think that’s a _very_ good idea and your concerns about this PR blowing up are very valid, too. We’re currently interested in the dashboard object, and not the wrapper. Any future additions should probably happen via new endpoints.

noopkat:
@tkoster I am so glad this has already made it in to production. AMAZING work! 👏 

thank you @kleinfreund for reviewing this thoroughly before I looked at it 🤘 

dan schiffman being the best and also saying "awesome"

PR #213, opened 2020-05-01
Feat/auto alex run on push

### LINKED ISSUE #212: Run Alex programmtically via pre-commit or something?
Looks like we have Alex hooked up, and that is great! But...

Who is running it, and how is it enforced?

mtheoryx:
Closes #212

The individual commits show:

1. Add the husky hook
2. Break the rule to show what WOULD have been caught
3. Fix the rule-breakage

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # electric-io
 
-The cutest IoT dashboard of your dreams ☁️
+Simply The cutest IoT dashboard of your dreams ☁️
 
 ![screenshot of electric io dashboard](screenshot.png)
 
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # electric-io
 
-Simply The cutest IoT dashboard of your dreams ☁️
+The cutest IoT dashboard of your dreams ☁️
 
 ![screenshot of electric io dashboard](screenshot.png)
 
--- a/package.json
+++ b/package.json
@@ -85,7 +85,8 @@
   },
   "husky": {
     "hooks": {
-      "pre-commit": "pretty-quick --staged"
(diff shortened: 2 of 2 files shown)


mtheoryx:
Also to clarify, I added this as a pre-push hook. I personally find pre-commit hooks to be sorta distracting. And it's not fun when you really do (for whatever reason) need to get something committed real quick, and you can't because of husky hooks.

kleinfreund:
@mtheoryx Indeed. It’s a worthwhile discussion to have: Where should linting take place in general? It starts at “no linting” and goes over “linting in continuous integration” to “linting on push or commit” and eventually all the way to “linting in the editor”. I certainly get the argument of linting getting in the way during the prototyping and/or implementation phase being a bad thing (see also the general TypeScript conundrum). Maybe for a project like this, it’s too much cognitive load to have this process be part of the implementation stage and should instead be handled layer (i.e. in CI). It’s also nice to not have have auto formatting changes in the diff people are meant to review (something I’m not always doing a good job with avoiding).

Maybe that’s a good topic to exchange opinions on in the next stream. :)

mtheoryx:
Personally, I feel that this should happen in CI, ideally.

But I figured this could also be a good educational check to show someone immediately how language could be changed, rather than a fire-and-forget and maybe not looking at it ever again after that.

It's worth some thoughts and discussion though!

noopkat:
thanks @mtheoryx for making this pre-push. I didn't actually realise that it was only at the CI stage which imo is a bit too late and would be frustrating to see, or as y'all discussed, easily missed failures at a later disconnected point.

PR #215, opened 2020-05-16
Replace color picker with its external package version

kleinfreund:
It’s been long overdue. I finally got my act together and bundled up our color picker into its own component and published it on NPM: https://www.npmjs.com/package/vue-accessible-color-picker. 🍉 🍇

## Changes

Replaces the color picker component with its external package version. After a thorough clean up, I bundled the whole color picker component into its own standalone package and published it on NPM.

Removes all the no longer needed files. The color conversions were not removed entirely because other components depend on a subset of their functionality.


noopkat:
So pleased to see this component become a real library. The vue ecosystem is now better off with your work in it. Yay accessibility! 🎉 

Thanks for your tireless work on this - I can appreciate just how much time went into this PR and all of the previous code before this PR 🌈 

some purple yetis cheering

Issue #210, opened 2020-04-25
Route tests fail when not simulating


tkoster:
I encountered the 404 issue that @noopkat had on the weekend running the unit tests. I believe this is what is happening:

When running the routes unit tests, `server.js` attempts to connect to the real IoT Hub if your `.env` file says so. The tests even print the `console.log` lines during hub startup. However, the _live hub_ fails to start and throws exceptions. I can reliably reproduce and fix the issue by switching between my `.env` file and the default `.env` file.

The environment affects the unit tests at runtime - we should aim to make the tests immune to the environment.

tkoster:
More info:

The `startHubServer` function initialises the hub before registering any routes with express (presumably for the good reason that the handlers need to be injected with the appropriate hub service). If initialising the hub throws an exception, the routes are not initialised.

A quick fix might be to completely ignore `.env` and force simulation mode when `NODE_ENV` is `test`, similar to how the HTTP server start is elided.

However, the root cause seems to be that `server.js` is not really a reusable module that one can "require" (as the tests do), but rather is the "main" of the Node app. What happens when you "require" it is quite nondeterministic: it changes its behaviour drastically depending on the environment.

A more permanent fix might be to reduce the nondeterminism in these functions by changing references to the environment variables into formal parameters that can be controlled by the tests, and then move the functions into a module that can be required by the tests without these side-effects.

tkoster:
Related to this: I noticed that the route tests also print lines saying an existing dashboard file was found and loaded. This is another source of potential nondeterminism that we should try to remove from the tests.

noopkat:
👍 💯 to all of this

tkoster:
I'll close this since #221 is merged and addresses this.

We can create a separate issue if dashboard files become a problem for tests. (I'm not convinced they currently are.)

PR #221, opened 2020-12-19
Ignore the user's .env file when running tests

tkoster:
Hi noopcrew! :wave:

I have been enjoying the avrgirl-arduino activities on the stream, even though I know hardly anything about microcontrollers, but I've been learning!

This is a small change for Electric I/O :zap: that addresses issue #210.

---

# The problem

When you run `npm run test` with `SIMULATING=false`, many of the route tests fail. The status code assertions fail because the web server returns 404s instead of the expected status codes. This is because the live hub fails to start, causing the express routes not to be added.

To reproduce the issue, first configure your `.env` file to disable the simulator and set an IoT Hub connection string. The connection string does not have to be for a real IoT Hub.

```
CONNECTION_STRING='HostName=xxxxxxxxx.azure-devices.net;SharedAccessKeyName=xxxxxxxxx;SharedAccessKey=xxxxxxxxx'
CONSUMER_GROUP=''
SIMULATING=false
```

Then run the tests:

```
$ npm run test
⋮
  ● API endpoints › HubService endpoints › can’t queue message on device if the service produces an error

    expect(received).toBe(expected) // Object.is equality

    Expected: 500
    Received: 404
⋮
```

# The fix

This change ignores the configuration in the user's `.env` file when running tests, that is, when `NODE_ENV === "test"`.

The reason this fixes the route tests is because it forces the use of the simulation hub (and ignore any potential future configuration in `.env` that might also reduce the determinism of the tests).

Fixing the tests to pass even when using the live hub could be an enhancement for the future.

### DIFF
--- a/lib/services/dashboard.service.js
+++ b/lib/services/dashboard.service.js
@@ -1,4 +1,3 @@
-require("dotenv").config();
 const debug = require("debug")("server");
 const fs = require("fs");
 const path = require("path");
--- a/server.js
+++ b/server.js
@@ -1,4 +1,9 @@
-require("dotenv").config();
+const dotenv = require("dotenv");
+
+if (process.env.NODE_ENV !== "test") {
+  dotenv.config();
+}
+
 const path = require("path");
 const http = require("http");
 const debug = require("debug")("server");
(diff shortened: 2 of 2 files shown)


mtheoryx:
This looks great to me!

PR #226, opened 2021-01-30
Update some dependencies (yeah, that’s right)

kleinfreund:
Updates some dependencies in the project. All vue-related dependencies were updated to versions not introducing breaking changes. All eslint-related dependencies were updated as far as possible.

Changes the husky pre-commit hook to just run the eslint command with the `--fix` argument and drops the pretty-quick and prettier packages (that’s right, they go in the square hole). They’ve always caused quite a bit of friction and the occassional conflict with the eslint configuration. All of the explicitly configured eslint rules were also removed in an attempt to make the whole topic of linting less stressful and annoying. Let’s see how this pans out.

(Following this, I have the changes for upgrading to webpack 5 ready in https://github.com/noopkat/electric-io/pull/227 which is branched-off of this PR’s branch.)

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,1 @@
+/public/js/dist
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -3,14 +3,5 @@
     "ecmaVersion": 9,
     "sourceType": "module"
   },
-  "rules": {
-    "semi": "error",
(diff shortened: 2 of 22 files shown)


noopkat:
thank you @kleinfreund ✨ 

spongebob being dusted by patrick

PR #227, opened 2021-01-30
Update bundling dependencies (upgrade to webpack 5)

kleinfreund:
*(This is a follow-up to https://github.com/noopkat/electric-io/pull/226 and should not be merged before it as it includes the commit of that dependency PR.)*

Updates all webpack and babel related dependencies.

Replaces the `node.fs === "empty"` webpack option with `resolve.fallback.fs === false`.

Moves the `devtool` webpack option to a ternary expression after defining the base configuration because `""` is no longer a valid value for the `devtool` option.

Adds the `.vue` file extension to all import statements.

Renames a deterministic bundle ID reference to whatever webpack now came up with. I haven’t got a single clue why it changed it in the first place.


noopkat:
all aboard the dependency update PR train 🥳 

cute animals dusting

PR #228, opened 2021-01-31
Update remaining development dependencies

kleinfreund:
*(This is a follow-up to #227 and should not be merged before it as it includes the commits of two dependency PRs.)*

Updates all remaining development dependencies.

Updates a bunch of Vue tests to make them pass a newly introduced axe rule.

Updates the year in the license file. 😁


noopkat:
love this chain of upgrades 🧹 

a rapidly moving cartoon person cleaning a house with dust spewing into the air

